PR #1, opened 2018-03-16
Use "use" statement instead of full namespace in code

mickaelandrieu:


dunglas:
Sorry I missed this one :(.
It's probably not needed anymore with the new API. WDYT?

mickaelandrieu:
I think we can close :) Can't wait to use it, this is something missing in PHP ecosystem!

dunglas:
It's not announced yet, and it's very experimental, the API will change again, but it works. Feedback and bug reports are very welcome!

PR #13, opened 2018-04-09
update composer command

coraxster:
there is no stable version now. Default composer errors with

[InvalidArgumentException]
  Could not find a version of package dunglas/panthere matching your minimum-stability (stable). Require it with an explicit version constraint allowing its desired stability.


dunglas:
Thanks @coraxster

PR #14, opened 2018-04-09
Native Firefox support (through geckodriver)

dunglas:
Allow to use Firefox as browser. The only requirement is to have Firefox installed.

Usage:

```php
$client = \Panthere\Client::createFirefoxClient();

// Use Panthère as usual
```

Or in a test case:

```php
$client = self::createPanthereClient(self::FIREFOX);
```

Needs facebook/php-webdriver#560.


src/ProcessManager/FirefoxManager.php
```diff
@@ -0,0 +1,69 @@
...
+
+        return RemoteWebDriver::create('http://localhost:4444', DesiredCapabilities::firefox(), null, null, null, null, null, true);
+    }
+
+    public function quit(): void
+    {
+        $this->process->stop();
+    }
+
+    private function findGeckodriverBinary(): string
```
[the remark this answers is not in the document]
dunglas: Why not, but you can already pass the path in the constructor: https://github.com/dunglas/panthere/pull/14/files#diff-c41515551a88c7160f9baa8ee31146caR30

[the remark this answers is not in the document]
Pierstoval: Yes, I saw that, but when testing the project or automating stuff, an env var is much more convenient for everyone without having to code the hack 😄

composer.json
```diff
@@ -16,9 +16,15 @@
       "homepage": "https://symfony.com/contributors"
     }
   ],
+  "repositories": [
+    {
+      "type": "vcs",
+      "url": "https://github.com/dunglas/php-webdriver"
```
stof: This VCS repo should be removed

README.md
```diff
@@ -85,6 +85,9 @@ This listener will start the web server on demand like previously, but it will s
 require __DIR__.'/vendor/autoload.php'; // Composer's autoloader
 
 $client = \Symfony\Component\Panther\Client::createChromeClient();
+// Or, if you care about the open web and prefer to use Firefox
```
Pierstoval: :heart:

src/DomCrawler/Crawler.php
```diff
@@ -173,7 +173,7 @@ public function attr($attribute): ?string
             return $this->text();
         }
 
-        return $element->getAttribute($attribute);
+        return $element->getAttribute($attribute) ?? '';
```
Pierstoval: Why not this?
```suggestion
        return $element->getAttribute($attribute) ?: '';
```

src/DomCrawler/Link.php
```diff
@@ -54,6 +55,6 @@ protected function setNode(\DOMElement $node)
 
     protected function getRawUri()
     {
-        return $this->element->getAttribute('href');
+        return $this->element->getAttribute('href') ?? '';
```
Pierstoval: Shouldn't it return the current URI?
```suggestion
        return $this->element->getAttribute('href') ?: $this->currentUri;
```

src/PantherTestCaseTrait.php
```diff
@@ -80,6 +80,7 @@ trait PantherTestCaseTrait
         'router' => '',
         'external_base_uri' => null,
         'readinessPath' => '',
+        'browser' => self::CHROME,
```
Pierstoval: Why not firefox as base? :fox_face:

src/DomCrawler/Crawler.php
dunglas: Because a ternary is less explicit? Another solution is just to cast `return (string) $element->getAttribute($attribute));` It's maybe more elegant.

src/PantherTestCaseTrait.php
dunglas: Because it's a huge BC break, and because Chrome is more used (unfortunately): 64% for Chrome, 5% for Firefox.

src/DomCrawler/Link.php
dunglas: Honestly, I don't know.

dunglas:
For an unknown reason, Travis uses an outdated version of `.travis.yml` for the PR build. It's why PR tests are red. As this branch is rebased against `master`, it's ok to merge as is.

src/DomCrawler/Crawler.php
Pierstoval: Totally agree (even though it's merged :p )

PR #18, opened 2018-04-15
Fix command-line arguments & allow using env vars and wildcard IP addresses for server

Pierstoval:
Using PHP 7.2.2 on Windows, testing this code in the WebServerManager:

```php
$commandLine = [$binary] + $finder->findArguments() + ['-dvariables_order=EGPCS', '-S', \sprintf('%s:%d', $this->hostname, $this->port)];

$commandLine2 = array_merge([$binary], $finder->findArguments(), ['-dvariables_order=EGPCS', '-S', \sprintf('%s:%d', $this->hostname, $this->port)]);

dump($commandLine);
dump($commandLine2);
```

And I have this output:

```
array:3 [
  0 => "E:\dev\php72-nts\php.exe"
  1 => "-S"
  2 => "0.0.0.0:12000"
]
array:4 [
  0 => "E:\dev\php72-nts\php.exe"
  1 => "-dvariables_order=EGPCS"
  2 => "-S"
  3 => "0.0.0.0:12000"
]
```

The first one is wrong, so I decided to refactor it with a plain `array_merge()` that will surely be working 👍 

This PR also introduces lots of other changes that allow using env vars and ips like `0.0.0.0` for servers listening to any host. I thought it would be a really nice improvement

### DIFF
--- a/src/ProcessManager/WebServerReadinessProbeTrait.php
+++ b/src/ProcessManager/WebServerReadinessProbeTrait.php
@@ -39,6 +39,13 @@ trait WebServerReadinessProbeTrait
     public function waitUntilReady(Process $process, string $url): void
     {
         $host = parse_url($url, PHP_URL_HOST);
+
+        if ($host === '0.0.0.0') {
+            // When server listens to any host, we can't ping 0.0.0.0 to check if server is ready.
+            // So we listen to local host to be sure it's accessible anyway.
+            $host = '127.0.0.1';
+        }
+
         $port = parse_url($url, PHP_URL_PORT);
 
         $retries = 0;
--- a/src/ProcessManager/WebServerReadinessProbeTrait.php
+++ b/src/ProcessManager/WebServerReadinessProbeTrait.php
@@ -40,7 +40,7 @@ trait WebServerReadinessProbeTrait
     {
         $host = parse_url($url, PHP_URL_HOST);
 
-        if ($host === '0.0.0.0') {
+        if ('0.0.0.0' === $host) {
             // When server listens to any host, we can't ping 0.0.0.0 to check if server is ready.
             // So we listen to local host to be sure it's accessible anyway.
             $host = '127.0.0.1';
@@ -70,7 +70,7 @@ trait WebServerReadinessProbeTrait
             \usleep(1000);
         } while (Process::STATUS_STARTED !== $status || ++$retries === $maxRetries);
 
-        if (count($socketErrors)) {
+        if (\count($socketErrors)) {
             throw new \RuntimeException(implode("\n", $socketErrors));
         }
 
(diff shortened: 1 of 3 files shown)


dunglas:
Can you provide the bug fix in a separate PR please?

src/ProcessManager/WebServerReadinessProbeTrait.php
```diff
@@ -36,23 +36,46 @@ private function checkPortAvailable(string $hostname, int $port, bool $throw = t
         }
     }
 
-    public function waitUntilReady(Process $process, string $url, bool $ignoreErrors = false): void
+    public function waitUntilReady(Process $process, string $url): void
     {
-        $context = \stream_context_create(['http' => [
```
dunglas: Why this change? It will not work if the `/` route isn't defined for instance.

src/ProcessManager/WebServerReadinessProbeTrait.php
```diff
@@ -36,23 +36,46 @@ private function checkPortAvailable(string $hostname, int $port, bool $throw = t
...
-            'ignore_errors' => $ignoreErrors,
-            'protocol_version' => '1.1',
-            'header' => ['Connection: close'],
-            'timeout' => 1,
-        ]]);
-
-        while (Process::STATUS_STARTED !== ($status = $process->getStatus()) || false === @\file_get_contents($url, false, $context)) {
+        $host = parse_url($url, PHP_URL_HOST);
+
+        if ($host === '0.0.0.0') {
```
dunglas: Please use yoda style.

src/ProcessManager/WebServerReadinessProbeTrait.php
```diff
@@ -36,23 +36,46 @@ private function checkPortAvailable(string $hostname, int $port, bool $throw = t
...
 
+            if ($errno !== 0) {
+                $socketErrors[] = "#$errno:$errstr";
+            }
+
             // block until the web server is ready
             \usleep(1000);
+        } while (Process::STATUS_STARTED !== $status || ++$retries === $maxRetries);
+
+        if (count($socketErrors)) {
```
dunglas: `\count()`

src/ProcessManager/WebServerReadinessProbeTrait.php
Pierstoval: Because it instead uses a socket, I thought it might be a much better idea

dunglas: I'm 👎 to use a socket. Using a HTTP client makes sure that the underlying API speaks HTTP. (A server supporting another protocol may already use this port).

Pierstoval:
As #28 is merged, the bugfix part is OK.

For the rest, we discussed with @dunglas on Slack and for now the socket part is not something we really need.

Let's close 😄

PR #20, opened 2018-04-15
Delete `$baseUri` on test teardown

Pierstoval:
Then $baseUri becomes unusable after class teardown, to avoid potential reuse


dunglas:
Thanks!

PR #22, opened 2018-05-11
Allowing customization of host and port for ChromeManager

aequasi:


src/ProcessManager/ChromeManager.php
```diff
@@ -28,11 +28,13 @@
...
     private $process;
     private $arguments;
+    private $options;
 
-    public function __construct(?string $chromeDriverBinary = null, ?array $arguments = null)
+    public function __construct(?string $chromeDriverBinary = null, ?array $arguments = null, ?array $options = null)
     {
         $this->process = new Process([$chromeDriverBinary ?? $this->findChromeDriverBinary()], null, null, null, null);
         $this->arguments = $arguments ?? $this->getDefaultArguments();
+        $this->options = $options ?? $this->getDefaultOptions();
```
dunglas: Can you use `array_merge` to merge data by the user provided and default values (in case it's not provided). The default value can then be `[]`.

src/ProcessManager/ChromeManager.php
```diff
@@ -89,4 +92,14 @@ private function getDefaultArguments(): array
...
         return $args;
     }
+
+    private function getDefaultOptions(): array
+    {
+        return [
+            'https' => false,
+            'host' => '127.0.0.1',
+            'port' => 9515,
+            'status' => '/status'
```
dunglas: `'path'` (again to be compatible with `parse_url`)?

src/ProcessManager/ChromeManager.php
```diff
@@ -89,4 +92,14 @@ private function getDefaultArguments(): array
 
         return $args;
     }
+
+    private function getDefaultOptions(): array
+    {
+        return [
+            'https' => false,
```
dunglas: `'scheme' => 'https' ` to respect the format of [`parse_url`](http://php.net/manual/en/function.parse-url.php) and to allow using potential exotic schemes?

src/ProcessManager/ChromeManager.php
aequasi: Yup! was just copying args

src/ProcessManager/ChromeManager.php
aequasi: Way better than what I had... :D

src/ProcessManager/ChromeManager.php
```diff
@@ -28,11 +28,13 @@
 
     private $process;
     private $arguments;
+    private $options;
 
-    public function __construct(?string $chromeDriverBinary = null, ?array $arguments = null)
+    public function __construct(?string $chromeDriverBinary = null, ?array $arguments = null, ?array $options = [])
```
dunglas: `?` should now be removed.

aequasi:
Not sure where this belongs in the readme. You have no examples of how to use the ChromeManager by itself.

aequasi:
Please let me know where you'd like me to put this @dunglas

dunglas:
Thanks! Regarding the documentation, what do you think of adding a new « Chrome » section in the README?

PR #23, opened 2018-06-01
Fix wrong statement in README

Oliboy50:
this seems wrong when you open the `composer.json` file
=> https://github.com/dunglas/panthere/blob/master/composer.json#L18-L19

## How?
Say that Panthère can run on every PHP project instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Panthère is very easy to use, because it implements the popular Symfony's [Brow
 [DomCrawler](https://symfony.com/doc/current/components/dom_crawler.html) APIs, , and contains
 all features you need to test your apps. It will sound familiar if you have ever created [a functional test for a Symfony app](https://symfony.com/doc/current/testing.html#functional-tests):
 as the API is exactly the same!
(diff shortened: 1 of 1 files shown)


dunglas:
Actually it depends of some Symfony Components, but not of the Symfony Framework. But 👍 for the change.

README.md
```diff
@@ -11,7 +11,7 @@ Panthère is very easy to use, because it implements the popular Symfony's [Brow
 [DomCrawler](https://symfony.com/doc/current/components/dom_crawler.html) APIs, , and contains
 all features you need to test your apps. It will sound familiar if you have ever created [a functional test for a Symfony app](https://symfony.com/doc/current/testing.html#functional-tests):
 as the API is exactly the same!
-Keep in mind that Panthère doesn't depend of Symfony, it's a standalone library.
+Keep in mind that Panthère could be used in every PHP project, it's a standalone library.
```
dunglas: s/could/can/ ?

dunglas:
Thanks @Oliboy50

PR #28, opened 2018-06-11
Fixed command-line that was not retrieving all arguments properly

dunglas:


dunglas:
Thanks for this fix @Pierstoval!

PR #30, opened 2018-06-13
Allow keeping the webserver & client active even after test teardown

Pierstoval:
By keeping the server enabled, we try to guarantee more performances (at least this should be benchmarked)

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,7 +171,7 @@ Unlike testing and web scraping libraries you're used to, Panther:
 When you use the Panther client, the web server running in background will be started at runtime and stopped at test's
 teardown.
 
-If you want to save a few performances and launch the server at PHPUnit startup, you can add the `ServerListener` to
+If you want to improve performances and launch the server at PHPUnit startup, you can add the `ServerListener` to
 your PHPUnit configuration:
 
 ```xml
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ your PHPUnit configuration:
 <!-- phpunit.xml.dist -->
 
     <listeners>
-        <listener class="Panthere\ServerListener" />
+        <listener class="Panther\ServerListener" />
     </listeners>
 ```
 
--- a/README.md
+++ b/README.md
@@ -166,9 +166,9 @@ Unlike testing and web scraping libraries you're used to, Panther:
 * supports custom [Selenium server](https://www.seleniumhq.org) installations
 * supports remote browser testing services including [SauceLabs](https://saucelabs.com/) and [BrowserStack](https://www.browserstack.com/)
 
-### Use the `ServerListener` to always have a running web server
+### Using the `ServerListener` to Always Have a Running Web Server
 
-When you use the Panthère client, the web server running in background will be started at runtime and stopped at test's
+When you use the Panther client, the web server running in background will be started at runtime and stopped at test's
 teardown.
 
 If you want to save a few performances and launch the server at PHPUnit startup, you can add the `ServerListener` to
@@ -178,7 +178,7 @@ your PHPUnit configuration:
 <!-- phpunit.xml.dist -->
 
     <listeners>
-        <listener class="Panther\ServerListener" />
+        <listener class="Symfony\Component\Panther\ServerListener" />
     </listeners>
 ```
 
--- a/README.md
+++ b/README.md
@@ -166,6 +166,24 @@ Unlike testing and web scraping libraries you're used to, Panther:
 * supports custom [Selenium server](https://www.seleniumhq.org) installations
 * supports remote browser testing services including [SauceLabs](https://saucelabs.com/) and [BrowserStack](https://www.browserstack.com/)
 
+### Use the `ServerListener` to always have a running web server
+
+When you use the Panthère client, the web server running in background will be started at runtime and stopped at test's
+teardown.
+
+If you want to save a few performances and launch the server at PHPUnit startup, you can add the `ServerListener` to
+your PHPUnit configuration:
+
+```xml
+<!-- phpunit.xml.dist -->
+
+    <listeners>
+        <listener class="Panthere\ServerListener" />
+    </listeners>
+```
+
+The listener will start the webserver when the test suite is started, and will stop it when all your tests are executed.
+
 ## Documentation
 
 Since Panther implements the API of popular libraries, it already has an extensive documentation:
--- a/src/ServerListener.php
+++ b/src/ServerListener.php
@@ -24,13 +24,13 @@ final class ServerListener implements TestListener
     public function startTestSuite(TestSuite $suite): void
     {
         echo "Starting Panther server for test suite {$suite->getName()}...\n";
-        PantherTestCase::stopServerOnTeardown();
-        PantherTestCase::startWebServer();
+        PantherTestCaseTrait::$stopServerOnTeardown = true;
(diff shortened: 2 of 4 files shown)


dunglas:
So when is the server stopped?

Pierstoval:
When the script ends?

Pierstoval:
I'm thinking about using a PHPUnit listener that would start the server at bootstrap and stop it when test suite has finished, and it would be opt-in (users would have to require it manually in their PHPUnit config, exactly like the PHPUnit Bridge listener in Symfony).

WDYT?

dunglas:
Good idea!

Pierstoval:
Added :wink:

tests/ServerListener.php
```diff
@@ -0,0 +1,40 @@
+<?php
```
dunglas: Should be in `src/` to be usable by the consumers, can you also use it in the project itself? It may speed out builds.

dunglas:
Can you also update the README file to explain how to use the new PHPUnit listener please?

src/PanthereTestCase.php
```diff
@@ -65,24 +66,29 @@
      */
     protected static $panthereClient;
 
-    public static function tearDownAfterClass()
+    /**
+     * @var bool
+     */
+    protected static $stopServerOnTeardown = true;
```
stof: should use `private` and `self::` to access it. The listener always call it on PanthereTestCase so it does not make sense to allow child classes to override the property to get separate storages.

Pierstoval: Yep, wanted to do it, seems like I forgot :sweat_smile: , nice catch thanks!

tests/ServerListener.php
Pierstoval: Done!

src/PanthereTestCase.php
Pierstoval: Done!

Pierstoval:
Rebased and updated the README :+1:

README.md
```diff
@@ -130,6 +130,24 @@ Unlike testing and web scraping libraries you're used to, Panthère:
 * supports custom [Selenium server](https://www.seleniumhq.org) installations
 * supports remote browser testing services including [SauceLabs](https://saucelabs.com/) and [BrowserStack](https://www.browserstack.com/)
 
+### Use the `ServerListener` to always have a running web server
+
+When you use the Panthère client, the web server running in background will be started at runtime and stopped at test's
```
dunglas: "Panther" now :D

README.md
```diff
@@ -130,6 +130,24 @@ Unlike testing and web scraping libraries you're used to, Panthère:
 * supports custom [Selenium server](https://www.seleniumhq.org) installations
 * supports remote browser testing services including [SauceLabs](https://saucelabs.com/) and [BrowserStack](https://www.browserstack.com/)
 
+### Use the `ServerListener` to always have a running web server
```
dunglas: Using the `ServerListener` to Always Have a Running Web Server

src/PanthereTestCaseTrait.php
```diff
@@ -52,7 +52,29 @@
      */
     protected static $panthereClient;
 
+    /**
+     * @var bool
+     */
+    private static $stopServerOnTeardown = true;
```
dunglas: Can't you use a public property instead? So we can get rid of the `stop` and `keep` methods?

src/ServerListener.php
```diff
@@ -0,0 +1,40 @@
...
+
+declare(strict_types=1);
+
+namespace Panthere;
+
+use PHPUnit\Framework\TestListener;
+use PHPUnit\Framework\TestListenerDefaultImplementation;
+use PHPUnit\Framework\TestSuite;
+
+class ServerListener implements TestListener
```
dunglas: `final`

src/ServerListener.php
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Panthere;
```
dunglas: `Symfony\Component\Panther` now

src/ServerListener.php
```diff
@@ -0,0 +1,40 @@
...
+class ServerListener implements TestListener
+{
+    use TestListenerDefaultImplementation;
+
+    public function startTestSuite(TestSuite $suite): void
+    {
+        echo "Starting Panthere server for test suite {$suite->getName()}...\n";
+        PanthereTestCase::stopServerOnTeardown();
+        PanthereTestCase::startWebServer(
+            getenv('PANTHERE_LISTENER_SERVER_DIR') ?: null,
```
dunglas: We use `$_SERVER` everywhere.

dunglas: And passing arguments should be avoided because env vars are already handled in the called method.

src/ServerListener.php
```diff
@@ -0,0 +1,40 @@
...
+            getenv('PANTHERE_LISTENER_HOSTNAME') ?: '127.0.0.1',
+            getenv('PANTHERE_LISTENER_PORT') ?: 9000
+        );
+    }
+
+    public function endTestSuite(TestSuite $suite): void
+    {
+        echo "\nShutting down Panthere server...\n";
```
dunglas: Panther

src/ServerListener.php
```diff
@@ -0,0 +1,40 @@
...
+use PHPUnit\Framework\TestListenerDefaultImplementation;
+use PHPUnit\Framework\TestSuite;
+
+class ServerListener implements TestListener
+{
+    use TestListenerDefaultImplementation;
+
+    public function startTestSuite(TestSuite $suite): void
+    {
+        echo "Starting Panthere server for test suite {$suite->getName()}...\n";
```
dunglas: Panther

dunglas:
This listener also needs to be tested.

src/PanthereTestCaseTrait.php
Pierstoval: It's called by the `ServerListener`, and I don't know if I can access the instantiated test from the listener :confused:

dunglas: You can keep the property static and call `PantherTestCase::$stopServerOnTeardown = false` right?

Pierstoval: Do you mean to make the property public?

README.md
```diff
@@ -166,6 +166,24 @@ Unlike testing and web scraping libraries you're used to, Panther:
 * supports custom [Selenium server](https://www.seleniumhq.org) installations
 * supports remote browser testing services including [SauceLabs](https://saucelabs.com/) and [BrowserStack](https://www.browserstack.com/)
 
+### Using the `ServerListener` to Always Have a Running Web Server
+
+When you use the Panther client, the web server running in background will be started at runtime and stopped at test's
+teardown.
+
+If you want to save a few performances and launch the server at PHPUnit startup, you can add the `ServerListener` to
```
javiereguiluz: `If you want to save a few performances ...` -> `If you want to improve performance ...`

dunglas:
A next step would be to keep Chromedriver working too, it will definitely improves the performance (in a new PR :D)?

dunglas:
Thank you very much @Pierstoval

PR #31, opened 2018-06-13
Explicitly add document root to PHP's web server command line

Pierstoval:
Else, on Windows, web-server is not created properly (possibly because working directory is not injected in the cmd wrapper, but not sure).

Here we are certain that the document root is correct all the time


dunglas:
Thanks!

PR #37, opened 2018-06-21
Fix installation instruction

voronkovich:


fabpot:
Good catch, thanks @voronkovich.

PR #42, opened 2018-06-29
 Add a blocker to avoid using HttpFoundation request and response objects

### LINKED ISSUE #17: Impossible to get the HttpFoundation objects from the Client
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | yes

When using `PanthereTestCase`, and when making a request, we can't get access to the `Symfony\Component\HttpFoundation\Response` object that should normally be sent to the Kernel. Instead, we only get access to the `BrowserKit` one.

Steps to reproduce:

```php
<?php

use Panthere\PanthereTestCase;

class MyPanthereTest extends PanthereTestCase
{
    public function testMyApp()
    {
        $client = static::createPanthereClient();
        $crawler = $client->request('GET', static::$baseUri.'/any-page');

        static::assertSame(200, $client->getResponse()->getStatusCode());
    }
}
```

PHPUnit's output:
```
Error : Call to undefined method Symfony\Component\BrowserKit\Response::getStatusCode()
```

I consider this as a BC break, because the `$client->request()` method SHOULD return an instance of the `Symfony\Component\BrowserKit\Client` class, for which `getResponse()` returns an instance of `Symfony\Component\HttpFoundation\Response`.

This is the same for the `Request` via `$client->getRequest()` by the way.

The issue comes from these lines:
https://github.com/dunglas/panthere/blob/b3f0601e105010d365360f238ca1b572a4e5ec82/src/Client.php#L238-L241

IMO, the Request and Response object should be either blocked from the PanthereClient via an exception (easy pick), or created & adapted based on the WebDriver results (harder).

WDYT?

Pierstoval:
Fixes #17

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -259,9 +259,9 @@ final class Client extends BaseClient implements WebDriver
             $uri = $this->baseUri.$uri;
         }
 
-        $this->request = $this->internalRequest = new Request($uri, 'GET');
+        $this->internalRequest = new Request($uri, 'GET');
         $this->webDriver->get($uri);
-        $this->response = $this->internalResponse = new Response($this->webDriver->getPageSource());
+        $this->internalResponse = new Response($this->webDriver->getPageSource());
         $this->crawler = $this->createCrawler();
 
         return $this;
--- a/src/Client.php
+++ b/src/Client.php
@@ -75,6 +75,16 @@ final class Client extends BaseClient implements WebDriver
         }
     }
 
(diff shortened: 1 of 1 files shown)


stof:
what about the getters for the BrowserKit request and response too ?

Pierstoval:
They actually are available and set by the  `Client::get()` method when making a request

dunglas:
Thanks!

Issue #41, opened 2018-06-29
Update example code with a simple css select and input text entry


ChangePlaces:
As a complete newbie, I think a quick addition to the example code will be useful to help populate inputs, so something like:

``` 
        // populate the search box with some text - this will trigger a search box to appear with results
        $client->findElement(WebDriverBy::cssSelector('.search__input.ds-input'))
            ->sendKeys('how');

        // Wait for the dynamic searchbox to appear
        $client->waitFor('#algolia-autocomplete-listbox-0');
```

PR #43, opened 2018-06-29
added example to populate text input

ChangePlaces:
Add form input manipulation and wait for results overlay to appear as per #41

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,13 @@ $crawler = $client->click($link);
 // Wait for an element to be rendered
 $client->waitFor('.support');
 
+ // populate the search box with some search text - this will trigger a search box overlay to appear with results
+$client->findElement(WebDriverBy::cssSelector('.search__input.ds-input'))
+    ->sendKeys('how');
+
+// Wait for the dynamic searchbox to appear
+$client->waitFor('#algolia-autocomplete-listbox-0');
+
 echo $crawler->filter('.support')->text();
 $client->takeScreenshot('screen.png'); // Yeah, screenshot!
 ```
--- a/examples/basic.php
+++ b/examples/basic.php
@@ -22,5 +22,12 @@ $crawler = $client->click($link);
 // Wait for an element to be rendered
 $client->waitFor('.support');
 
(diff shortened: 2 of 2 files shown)


ChangePlaces:
I added the code to the example file also

README.md
```diff
@@ -42,6 +42,13 @@ $crawler = $client->click($link);
 // Wait for an element to be rendered
 $client->waitFor('.support');
 
+ // populate the search box with some search text - this will trigger a search box overlay to appear with results
+$client->findElement(WebDriverBy::cssSelector('.search__input.ds-input'))
+    ->sendKeys('how');
```
javiereguiluz: Just asking: is mandatory to use `sendKeys()` as the method name because it comes from some external library ... or did we define this method ourselves? I'm asking because "send keys" looks like taking the keys out of your keyboard, putting them in a box and sending them somewhere. The word that we're looking for is probably `type`.

Also, a word about the DX of the proposed API: it looks too low-level and verbose. Example:

```php
// Panthere
$client->findElement(WebDriverBy::cssSelector('.search__input.ds-input'))->sendKeys('how');

// Laravel's Dusk
$browser->type('.search__input.ds-input', 'how');

// it's also possible to find elements by name, not only CSS selector
// $browser->type('search', 'how');
```

stof: The return value of `findElement` is a WebDriverElement, i.e. something coming from an external lib

dunglas: Indeed, and the Facebook's lib just map W3C/Selenium terms. However, we should be able to use the BrowserKit API directly:

```php
$form = $client->filter('.search form')->form();
$form['search']->setValue('my search terms');

$client->submit($form);
```

I've not tested it, but if it works with BrowserKit, it works with Panthère).

Anyway, we can add new helper methods, but we should provide them in BrowserKit directly (the idea behind Panthère is too be able to use BrowserKit or Goutte or a real browser with the exact same code, and maybe some conditions for the waits for instance).

ChangePlaces:
Maybe the discussion for modifying the function names should be put into a new issue?

dunglas:
Closing as it's now possible to just use the standard BrowserKit methods.

Issue #40, opened 2018-06-29
Should the composer install instruction be --dev?


ChangePlaces:
e.g. `composer require --dev symfony/panthere:dev-master`

ChangePlaces:
I agree!

PR #44, opened 2018-06-29
Added quick note about using the --dev flag

ChangePlaces:
Added quick note about using the --dev flag in the install instructions as per #40


dunglas:
Thanks @ChangePlaces

PR #48, opened 2018-07-06
Make update.sh compatible with sh

Pierstoval:
Else, the update can't be made on non-bash shells (such as Alpine for example).

IMO it's better to stick to `sh` which is available almost everywhere


dunglas:
👎 this tools is intended for maintainers only, and bash provide better primitives than just sh (arrays and maps for instance).

dunglas:
Closing for now

stof:
@dunglas what about excluding this file using `.gitattributes`, so that non-maintainers are less likely to see it ?

dunglas:
@stof good idea, indeed

PR #49, opened 2018-07-06
Allow using an env var to determine chrome driver binary

Pierstoval:
My problem is that Panthère doesn't work on Alpine (because on Alpine we should either use the driver coming from the repository, or recompile it manually).

For this, it could be nice to tell Alpine users to use either the package or to recompile it instead of using Panthère's provided binary. And in general, to get the latest version, it's better to propose them to download the package in another way, for them to be sure 😅.

For now, an env var is the best solution to use another binary, to me. 

WDYT?

### DIFF
--- a/src/ProcessManager/ChromeManager.php
+++ b/src/ProcessManager/ChromeManager.php
@@ -32,7 +32,7 @@ final class ChromeManager implements BrowserManagerInterface
 
     public function __construct(?string $chromeDriverBinary = null, ?array $arguments = null, array $options = [])
     {
-        $this->process = new Process([$chromeDriverBinary ?? $this->findChromeDriverBinary()], null, null, null, null);
+        $this->process = new Process([$chromeDriverBinary ?: $this->findChromeDriverBinary()], null, null, null, null);
         $this->arguments = $arguments ?? $this->getDefaultArguments();
         $this->options = \array_merge($this->getDefaultOptions(), $options);
     }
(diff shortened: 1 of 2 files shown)


src/ProcessManager/ChromeManager.php
```diff
@@ -32,7 +32,7 @@
 
     public function __construct(?string $chromeDriverBinary = null, ?array $arguments = null, array $options = [])
     {
-        $this->process = new Process([$chromeDriverBinary ?? $this->findChromeDriverBinary()], null, null, null, null);
+        $this->process = new Process([$chromeDriverBinary ?: $this->findChromeDriverBinary()], null, null, null, null);
```
dunglas: Why this change?

src/ProcessManager/ChromeManager.php
```diff
@@ -66,6 +66,10 @@ public function quit(): void
 
     private function findChromeDriverBinary(): string
     {
+        if (false !== $binary = \getenv('PANTHERE_CHROME_DRIVER_BINARY')) {
```
dunglas: we use `$_ENV` in other places (`getenv` isn't thread safe).

src/ProcessManager/ChromeManager.php
Pierstoval: To avoid `$chromeDriverBinary` be an empty string 😄

src/ProcessManager/ChromeManager.php
Pierstoval: Fixed

src/ProcessManager/ChromeManager.php
dunglas: How can it be?

Pierstoval: With a simple developer mistake, I know I might be paranoid, but if you think I should revert this, feel free to ask, and I will

dunglas: No it's ok! Can you just document this new env var in the README file please?

dunglas:
Thanks @Pierstoval

PR #51, opened 2018-07-07
Allow injecting env vars to web server process

Pierstoval:
After I closed #18 I forgot to make another PR for this part.

I think it's essential to automatically inject env vars.

As we are using PHPUnit, env vars for the test execution should be set in `phpunit.xml.dist`, therefore they'll be available in `$_ENV` thanks to PHPUnit, so we can safely inject them in the process.

This will avoid it use the `.env` file when using the full-stack setup, and make sure we can isolate test env from dev env when using the same machine for both.

### DIFF
--- a/src/PanthereTestCaseTrait.php
+++ b/src/PanthereTestCaseTrait.php
@@ -71,8 +71,7 @@ trait PanthereTestCaseTrait
         self::$baseUri = null;
     }
 
-    protected static function startWebServer(?string $webServerDir = null, string $hostname = '127.0.0.1', int $port = 9000): void
-    {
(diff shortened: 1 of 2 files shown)


Pierstoval:
Investigating on the issue 😉 

Then this PR becomes an RFC just to propose the end-user to add and/or overwrite env vars in the current web server! 😄

dunglas:
I've found the underlying issue, and I fixed the Flex recipe accordingly: https://github.com/symfony/recipes/pull/464

Thank you very much for reporting this Alex!

Pierstoval:
:+1:

PR #53, opened 2018-07-09
Use Symfony as root namespace

dunglas:
`Panthere\` becomes `Symfony\Component\Panthere\`


stof:
PR description should be updated though, to describe the right namespace

PR #56, opened 2018-07-20
fixed waitFor return type error

andrewmy:
In browser tests I got an exception:

> TypeError : Return value of Symfony\Component\Panthere\Client::waitFor() must be an instance of Symfony\Component\Panthere\object, instance of Facebook\WebDriver\Remote\RemoteWebElement returned

Of course, `\object` and `\stdClass` didn't pass as return types, so it appears the method should be without one. Come to think about it, can it return anything other than `RemoteWebElement`?

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -226,7 +226,7 @@ final class Client extends BaseClient implements WebDriver
         return new CookieJar($this->webDriver);
     }
 
-    public function waitFor(string $cssSelector, int $timeoutInSecond = 30, int $intervalInMillisecond = 250): object
+    public function waitFor(string $cssSelector, int $timeoutInSecond = 30, int $intervalInMillisecond = 250)
     {
         return $this->wait($timeoutInSecond, $intervalInMillisecond)->until(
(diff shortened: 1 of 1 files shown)


stof:
Which version of PHP are you using ? the `object` typehint should work fine on PHP 7.2+

andrewmy:
```
PHP 7.2.7 (cli) (built: Jun 22 2018 06:27:50) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.2.0, Copyright (c) 1998-2018 Zend Technologies
    with Xdebug v2.6.0, Copyright (c) 2002-2018, by Derick Rethans
    with Zend OPcache v7.2.7, Copyright (c) 1999-2018, by Zend Technologies
```

stof:
@andrewmy are you sure this is the PHP runtime used when running your code ? The error message you got is exactly the one seen when running such code on PHP 7.0 or PHP 7.1): https://3v4l.org/Um2cs

andrewmy:
Indeed, for some reason running `./bin/phpunit` in the project dir picked up the wrong runtime.
Thanks for pointing this out, closing the PR.

PR #70, opened 2018-07-26
Fixes #69 fixing scrapping typos

### LINKED ISSUE #69: typo in packagist description
> A browser testing and web __scrapping__ library for PHP and Symfony.

scrapping should be "scraping"

ChangePlaces:
as requested #69

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,8 +1,8 @@
 {
   "name": "symfony/panther",
   "type": "library",
-  "description": "A browser testing and web scrapping library for PHP and Symfony.",
-  "keywords": ["scrapping", "E2E", "testing", "webdriver", "selenium", "symfony"],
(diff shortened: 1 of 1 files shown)


ChangePlaces:
I don't think the error " Ignored error pattern /Parameter #1 \$value of method Symfony\\Component\\DomCrawler\\Field\\ChoiceFormField::select\(\) expects string, array<int, string> given\./ was not matched in reported errors. " is anything to do with this pr!

stof:
yeah, that ignore rule is not needed anymore, because the upstream phpdoc has been fixed in https://github.com/symfony/symfony/pull/27849

ChangePlaces:
I haven't touched any files other than the `composer.json`, so I surely don't have to pull anything? Can this not be merged without all green?

dunglas:
Thanks @ChangePlaces!

PR #82, opened 2018-09-13
Add Symfony 3.4 support

### LINKED ISSUE #79: Symfony 3.4 support
Hi,

is there a specific reason why Panther requires `symfony/browser-kit` and `symfony/console` in `^4.0` version?

Since Symfony 4.0 does not yet have an LTS version, and since it is fully compatible with 3.4, it would be nice to allow installing Panther with Symfony 3.4 present.

Thanks!

emodric:
Fixes #79

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,13 +10,23 @@ cache:
   directories:
     - $HOME/.composer/cache
 
+matrix:
+  include:
+    -
+      env: DEPS="low"
+
 before_install:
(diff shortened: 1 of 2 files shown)


emodric:
@dunglas How about this?

dunglas:
Well, it looks like Panther doesn't work with 3.4.

emodric:
Hah, I guess so!

Then this obviously can be closed :)

PR #84, opened 2018-09-17
Bump min PHP version up to 7.2

y2keeper:
Fixes #39 - Panther uses PHP 7.2 language features such as object return type and should therefore require a min PHP version of 7.2

PR #86, opened 2018-09-18
Fix PHP 7.1 compat

dunglas:
Closes #84

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -15,6 +15,7 @@ namespace Symfony\Component\Panther;
 
 use Facebook\WebDriver\Exception\NoSuchElementException;
 use Facebook\WebDriver\Exception\TimeOutException;
+use Facebook\WebDriver\Remote\RemoteWebElement;
 use Facebook\WebDriver\WebDriver;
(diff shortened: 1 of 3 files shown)


src/Client.php
```diff
@@ -226,7 +228,13 @@ public function getCookieJar()
         return new CookieJar($this->webDriver);
     }
 
-    public function waitFor(string $cssSelector, int $timeoutInSecond = 30, int $intervalInMillisecond = 250): object
+    /**
+     * @throws NoSuchElementException
+     * @throws TimeOutException
+     *
+     * @return object
```
stof: shouldn't this say it returns a WebDriverElement instead ? `->until` is documented as returning the return value of the callable. If you just rely on that, the return type would be `mixed`, not `object`. And if you take the callable into account, you can be more precise than `object` (which is not much better than `mixed` as far as autocompletion is concerned)

stof:
this should also add a Travis job running on PHP 7.1 to ensure we don't reintroduce another incompatibility.

stof:
hmm, actually, we already have a CI job running on PHP 7.1. so what we actually need is a test using this API.

PR #96, opened 2018-09-20
Ability to set a custom "google-chrome" binary

Hakadel:
I added the ability to set a custom Chrome binary by using "GOOGLE_CHROME_SHIM" environment variable.

You need it to be able to use Panther with Heroku (https://elements.heroku.com/buildpacks/heroku/heroku-buildpack-google-chrome)

Otherwise you'll have the "cannot find Chrome binary" error.

### DIFF
--- a/src/ProcessManager/ChromeManager.php
+++ b/src/ProcessManager/ChromeManager.php
@@ -54,7 +54,7 @@ final class ChromeManager implements BrowserManagerInterface
             $chromeOptions = new ChromeOptions();
             $chromeOptions->addArguments($this->arguments);
             $capabilities->setCapability(ChromeOptions::CAPABILITY, $chromeOptions);
-            if (isset($_SERVER['PANTHER_CHROME_BINARY']) {
+            if (isset($_SERVER['PANTHER_CHROME_BINARY'])) {
                 $chromeOptions->setBinary($_SERVER['PANTHER_CHROME_BINARY']);
             }
         }
--- a/src/ProcessManager/ChromeManager.php
+++ b/src/ProcessManager/ChromeManager.php
@@ -53,6 +53,7 @@ final class ChromeManager implements BrowserManagerInterface
         if ($this->arguments) {
             $chromeOptions = new ChromeOptions();
             $chromeOptions->addArguments($this->arguments);
+            $chromeOptions->setBinary(getenv('PANTHER_CHROME_BINARY') ?: '');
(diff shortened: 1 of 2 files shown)


stof:
I suggest using `GOOGLE_CHROME_BIN` as env var. It will be more logical IMO (your alternate path may not be a shim). And it will stay compatible with the heroku buildpack you linked, as it sets both variables.

Hakadel:
@stof When I'm using the `GOOGLE_CHROME_BIN` env var, I got this error on Heroku : 

`app.CRITICAL: unknown error: Chrome failed to start: crashed   (unknown error: DevToolsActivePort file doesn't exist)   (The process started from chrome location /app/.apt/opt/google/chrome/chrome is no longer running, so ChromeDriver is assuming that Chrome has crashed.)`

This is why I used the `SHIM`

dunglas:
May I suggest `PANTHER_CHROME_BINARY` for consistency with other vars?
Then you'll be able to Panther on Heroku with: `PANTHER_CHROME_BINARY=$GOOGLE_CHROME_SHIM phpunit`

src/ProcessManager/ChromeManager.php
```diff
@@ -53,6 +53,7 @@ public function start(): WebDriver
         if ($this->arguments) {
             $chromeOptions = new ChromeOptions();
             $chromeOptions->addArguments($this->arguments);
+            $chromeOptions->setBinary(getenv('GOOGLE_CHROME_SHIM') ?: '');
```
dunglas: Can you allow to set this value with a new optional constructor parameter that will fallback to the environment variable please?

dunglas: Can you also use `$_SERVER['GOOGLE_CHROME_SHIM']` instead of `getenv`?

Gemorroj: @Hakadel ping

Hakadel: Sorry I was busy :x Do you think I should use $_SERVER['PANTHER_CHROME_BINARY'] instead of getenv ??? @Gemorroj @stof 

I'm gonna rebase soon as well.

Hakadel:
@stof I rebased the branch and used "PANTHER_CHROME_BINARY" to be consistent. What do you think?

dunglas:
Thanks @Hakadel!

PR #99, opened 2018-09-23
Automatically boot the SF kernel

dunglas:
Without this patch, using `self::$container` doesn't work.


dunglas:
I don't understand why Travis choose not the `master` branch as remote. But tests are green.

src/PantherTestCaseTrait.php
```diff
@@ -88,17 +89,27 @@ protected static function startWebServer(?string $webServerDir = null, string $h
...
+     * @param array $kernelOptions An array of options to pass to the createKernel method
+     */
+    protected static function createPantherClient(string $hostname = '127.0.0.1', int $port = 9000, array $kernelOptions = []): PantherClient
     {
         self::startWebServer(null, $hostname, $port);
         if (null === self::$pantherClient) {
             self::$pantherClient = Client::createChromeClient(null, null, [], self::$baseUri);
         }
 
+        if (\is_a(self::class, WebTestCase::class, true)) {
```
nicolas-grekas: `if (\is_a(self::class, KernelTestCase::class, true)) {` instead?

PR #100, opened 2018-09-24
Added .gitattributes

alexmanno:
Added files to .gitattributes

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,11 @@
+/examples/ export-ignore
+/appveyor.yml export-ignore
+/panther.png export-ignore
+/.php_cs.dist export-ignore
+/phpstan.neon export-ignore
+/phpunit.xml.dist export-ignore
+/tests/ export-ignore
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -0,0 +1,11 @@
...
+/appveyor.yml export-ignore
+/panther.png export-ignore
+/.php_cs.dist export-ignore
+/phpstan.neon export-ignore
+/phpunit.xml.dist export-ignore
+/tests/ export-ignore
+/.travis.yml export-ignore 
+/.gitignore export-ignore
+/.gitattributes export-ignore
+/README.md export-ignore
```
dunglas: We should kip it in the generated archive don't you think?

alexmanno: Do you mean the README.md?

dunglas: yes

alexmanno: If you prefer, I can remove the README from the ignored files

dunglas:
Thanks @alexmanno

PR #101, opened 2018-09-25
Add PHPDoc to update return type of inherited getCrawler method

ldawkes:
Panther's crawler client inherits the getCrawler method from browser-kit/client, can fix by adding a PHPDoc "method" parameter to the class definition.


dunglas:
Thanks @ldawkes!

PR #102, opened 2018-09-25
Add Client->refreshCrawler() method

ldawkes:
WebDriver currently doesn't refresh the crawler component when a page uses deferred redirects through JavaScript. 

TestCase's would need access to a refreshed Crawler component when testing anything asynchronous like a login form that consumes an API endpoint.


dunglas:
Thanks @ldawkes!

PR #104, opened 2018-09-27
Keep alive the webserver between test suites

maidmaid:
cf https://phpunit.readthedocs.io/en/7.3/extending-phpunit.html#extending-the-testrunner

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,23 +166,34 @@ Unlike testing and web scraping libraries you're used to, Panther:
 * supports custom [Selenium server](https://www.seleniumhq.org) installations
 * supports remote browser testing services including [SauceLabs](https://saucelabs.com/) and [BrowserStack](https://www.browserstack.com/)
 
-### Using the `ServerListener` to Always Have a Running Web Server
+### Improve Performances by Having a Persistent Web Server Running
 
-When you use the Panther client, the web server running in background will be started at runtime and stopped at test's
-teardown.
+When you use the Panther client, the web server running in background will be started on demand at the first call to
+`createPantherClient()`, `createGoutteClient()` or `startWebServer()` and it will be stopped at `tearDownAfterClass()`.
 
-If you want to improve performances and launch the server at PHPUnit startup, you can add the `ServerListener` to
-your PHPUnit configuration:
+If you want to improve performances, you can hook to PHPUnit in your `phpunit.xml.dist` configuration file with the
+Panther's server extension:
 
 ```xml
 <!-- phpunit.xml.dist -->
+    <extensions>
+        <extension class="Symfony\Component\Panther\ServerExtension" />
+    </extensions>
+```
 
+This extension will start the web server on demand like previously, but it will stop it after the very last test.
+
+It should be noted that the Panther's extension only works with PHPUnit `>= 7.3`. Nonetheless, if you are using an
+anterior PHPUnit version, you can also hook to PHPUnit with the Panther's server listener:
+
+```xml
+<!-- phpunit.xml.dist -->
     <listeners>
         <listener class="Symfony\Component\Panther\ServerListener" />
     </listeners>
 ```
 
-The listener will start the webserver when the test suite is started, and will stop it when all your tests are executed.
+This listener will start the web server on demand like previously, but it will stop it after each test suite.
 
 ## Documentation
 
--- a/src/PantherTestCaseTrait.php
+++ b/src/PantherTestCaseTrait.php
@@ -58,7 +58,7 @@ trait PantherTestCaseTrait
 
     public static function tearDownAfterClass()
     {
-        if (true === self::$stopServerOnTeardown) {
+        if (self::$stopServerOnTeardown) {
             static::stopWebServer();
         }
     }
--- a/src/ServerExtension.php
+++ b/src/ServerExtension.php
@@ -13,21 +13,22 @@ declare(strict_types=1);
 
 namespace Symfony\Component\Panther;
 
-use PHPUnit\Framework\TestListener;
-use PHPUnit\Framework\TestListenerDefaultImplementation;
-use PHPUnit\Framework\TestSuite;
+use PHPUnit\Runner\AfterLastTestHook;
+use PHPUnit\Runner\BeforeFirstTestHook;
 
-final class ServerListener implements TestListener
+/**
+ *  @author Dany Maillard <danymaillard93b@gmail.com>
+ */
+final class ServerExtension implements BeforeFirstTestHook, AfterLastTestHook
 {
-    use TestListenerDefaultImplementation;
     use ServerTrait;
 
-    public function startTestSuite(TestSuite $suite): void
+    public function executeBeforeFirstTest(): void
     {
         $this->keepServerOnTeardown();
     }
 
-    public function endTestSuite(TestSuite $suite): void
+    public function executeAfterLastTest(): void
     {
         $this->stopWebServer();
     }
--- a/src/ServerTrait.php
+++ b/src/ServerTrait.php
@@ -1,0 +1,32 @@
+<?php
+
+/*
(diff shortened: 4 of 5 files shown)


dunglas:
This should be the default suggestion, don't you think @Pierstoval?

maidmaid:
We could recommend to use the Extension for the PHPUnit users >= 7.3 and the Listener for the others ?

dunglas:
Good idea.

README.md
```diff
@@ -166,7 +166,7 @@ Unlike testing and web scraping libraries you're used to, Panther:
 * supports custom [Selenium server](https://www.seleniumhq.org) installations
 * supports remote browser testing services including [SauceLabs](https://saucelabs.com/) and [BrowserStack](https://www.browserstack.com/)
 
-### Using the `ServerListener` to Always Have a Running Web Server
+### Using the `ServerListener`/`ServerExtension` to Always Have a Running Web Server
```
Pierstoval: If we want to recommend 2 tools for different versions, I'd update this for the following:

```
Improve performances by having a persistent web server running
```

src/ServerExtension.php
```diff
@@ -0,0 +1,33 @@
...
+use PHPUnit\Runner\AfterLastTestHook;
+use PHPUnit\Runner\BeforeFirstTestHook;
+
+final class ServerExtension implements BeforeFirstTestHook, AfterLastTestHook
+{
+    public function executeBeforeFirstTest(): void
+    {
+        echo "Starting Panther server...\n";
+        PantherTestCaseTrait::$stopServerOnTeardown = true;
+        PantherTestCaseTrait::startWebServer();
```
Pierstoval: Instead of duplicating code, can we use a trait for these 3 lines and the ones from the other method, and then use the trait in both the `ServerExtension` class and `ServerListener`?

Pierstoval:
Very nice idea, I didn't know about this, else I would've done it already :sweat_smile: Thanks!

README.md
maidmaid: Any suggestion to reword the body of this chapter ?

Pierstoval: Yes:
(sorry couln't use a diff because I don't know how to handle code blocks inside code blocks :rofl: )

---

When you use the Panther client, the web server running in background will be started at runtime and stopped at test's
teardown.

If you want to improve performances and launch the server at test suite startup, you can hook to PHPUnit to
your `phpunit.xml.dist` configuration file with Panther's listeners:

```xml
<!-- phpunit.xml.dist -->

    <!-- PHPUnit <7.3 -->
    <listeners>
        <listener class="Symfony\Component\Panther\ServerListener" />
    </listeners>

    <!-- PHPUnit 7.3+ -->
    <extensions>
        <extension class="Symfony\Component\Panther\ServerExtension" />
    </extensions>
```

The listener will start the webserver when the test suite is started, and will stop it when all your tests are executed.

src/ServerExtension.php
maidmaid: Done.

README.md
maidmaid: Added.

src/ServerTrait.php
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * @author Dany Maillard <danymaillard93b@gmail.com>
+ *
+ * @internal
+ */
+trait ServerTrait
+{
+    private function keepServerOnTeardown(): void
+    {
+        PantherTestCase::$stopServerOnTeardown = false;
```
maidmaid: I removed `startWebServer()` here because if there is in the user tests something like :
```php
public static function setUpBeforeClass()
{
    static::createPantherClient(['webServerDir' => '.');
}
```
The config `webServerDir` will not be used.

maidmaid: Second point : calling the static property/method from the trait doesn't work, but from `PantherTestCase`, it works.

maidmaid:
Could you review again, I made some commented changes.

src/PantherTestCaseTrait.php
```diff
@@ -58,7 +58,7 @@
 
     public static function tearDownAfterClass()
     {
-        if (true === self::$stopServerOnTeardown) {
+        if (self::$stopServerOnTeardown) {
```
Pierstoval: Why did you make this change? IIRC it's always faster to explicitly compare values strictly instead of relying on the default PHP type-cast, WDYT?

src/ServerTrait.php
Pierstoval: Anyone using the listener/extension **must** remember that the server is **already running**, therefore stop it before starting a new one, so to me the server must be started anyway in the listener, it'll just be kept on teardown. But generally, changing server config between tests is not very common, the only case I might see is when one changes environment vars, but in such case, the listener would probably be a bad practice...

src/PantherTestCaseTrait.php
stof: Well, `true === self::$stopServerOnTeardown` just produces another zval, which is then checked again with non-strict rules in the `if`. So if this property always contains a boolean already, you are only adding *more* work with the `true ===`

src/ServerTrait.php
maidmaid: My point was more about how to configure the web server : in your case, if we want to use the listener/extension, the only way to configure the web service would be only through the `PANTHER_WEB_SERVER_*` env vars, and I would like to allow also the configuration through `createPantherClient($option)`.

src/PantherTestCaseTrait.php
maidmaid: If the gain is not significant as explained by @stof, would you be ok with this change ?

src/ServerTrait.php
maidmaid: Moreover, the listener/extension is working for all the user tests. So, if we want to run a `phpunit --filter UnitTest\\`, the web server will be launch whereas it's useless in this case.

src/PantherTestCaseTrait.php
Pierstoval: I wasn't aware of this zval process making another check, TIL :) 
I'm ok :)

src/ServerTrait.php
Pierstoval: Using the listener/extension shouldn't really be fully compatible with manually configuring the server IMO :confused:

maidmaid:
I edited the PR; now, when we use the extension:
- the web server and chrome are started through the first call to `createPantherClient()`, `createGoutteClient()` or of course `startWebServer()`,
- while the tests, the web server and chrome are kept alive,
- and they are stopped after the last test.

Thus, if we filter the tests for which there is no need of the web server and chrome, they are simply not launch.

src/ServerTrait.php
maidmaid: Let's continue this topic after my last comment https://github.com/symfony/panther/pull/104#issuecomment-425613384 :)

README.md
```diff
@@ -166,20 +166,26 @@ Unlike testing and web scraping libraries you're used to, Panther:
...
 
-### Using the `ServerListener` to Always Have a Running Web Server
+### Improve Performances by Having a Persistent Web Server Running
 
 When you use the Panther client, the web server running in background will be started at runtime and stopped at test's
 teardown.
 
-If you want to improve performances and launch the server at PHPUnit startup, you can add the `ServerListener` to
-your PHPUnit configuration:
+If you want to improve performances and launch the server at test suite startup, you can hook to PHPUnit to
```
Pierstoval: This part must be reworded then :smiley:

Pierstoval:
I actually like the idea of starting the server manually & keeping it, I don't know why I didn't do it in the first place :smile: 

Once readme is a bit reworded, everything will be ok :+1:

maidmaid:
Rewording pushed. I'm waiting for your feedback :)

dunglas:
Thanks @maidmaid!

PR #107, opened 2018-09-27
Client::waitFor() returns a crawler

maidmaid:
In addition to #102, I think that `Client::waitFor()` should directly return an updated crawler.


maidmaid:
Could you check again ? It looks like unrelated now.

dunglas:
Juste execute `php-cs-fixer fix` and it will be alright!

maidmaid:
Done :)

dunglas:
Thanks Dany!

PR #109, opened 2018-10-01
Fix hostname option

maidmaid:


maidmaid:
Oups, sorry for that :)

dunglas:
Thanks @maidmaid!

PR #112, opened 2018-10-03
Replace $_SERVER by $_ENV

maidmaid:
PHPUnit internally uses `$_ENV` for the env vars. This, it is possible to edit env vars through the `phpunit.xml.dist `:

```xml
<phpunit>
    <php>
        <env name="PANTHER_WEB_SERVER_ROUTER" value="router.php"/>
    </php>
</phpunit>
```
cf https://phpunit.readthedocs.io/en/7.3/configuration.html?highlight=env#setting-php-ini-settings-constants-and-global-variables

### DIFF
--- a/src/PantherTestCaseTrait.php
+++ b/src/PantherTestCaseTrait.php
@@ -102,10 +102,10 @@ trait PantherTestCaseTrait
         }
 
         $options = [
-            'webServerDir' => $options['webServerDir'] ?? static::$webServerDir ?? $_SERVER['PANTHER_WEB_SERVER_DIR'] ?? self::$defaultOptions['webServerDir'],
+            'webServerDir' => $options['webServerDir'] ?? static::$webServerDir ?? getenv('PANTHER_WEB_SERVER_DIR') ?: self::$defaultOptions['webServerDir'],
             'hostname' => $options['webServerDir'] ?? self::$defaultOptions['hostname'],
-            'port' => (int) ($options['port'] ?? $_SERVER['PANTHER_WEB_SERVER_PORT'] ?? self::$defaultOptions['port']),
-            'router' => $options['router'] ?? $_SERVER['PANTHER_WEB_SERVER_ROUTER'] ?? self::$defaultOptions['router'],
+            'port' => (int) ($options['port'] ?? getenv('PANTHER_WEB_SERVER_PORT') ?: self::$defaultOptions['port']),
+            'router' => $options['router'] ?? getenv('PANTHER_WEB_SERVER_ROUTER') ?: self::$defaultOptions['router'],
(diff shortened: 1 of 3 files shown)


stof:
but this might break when using actual env variables depending on the php.ini config (as `$_ENV` might not be populated)

maidmaid:
PHPUnit actually uses `putenv()` (https://github.com/sebastianbergmann/phpunit/blob/4f44ade4cf58e1fcddf68b3db8467fe1694552ea/src/Util/Configuration.php#L595). So, should I use `getenv()` instead `$_ENV` ?

maidmaid:
`getenv()` used in the last version :)

dunglas:
`getenv` isn't thread safe. I don't know if it can be an issue here (ping @nicolas-grekas).

nicolas-grekas:
Phpunit reads $_ENV and @dunglas is correct, no need to use getenv()

maidmaid:
getenv() properly returns env var from phpunit.xml.dist and from export. It's not the case with $_ENV :

```
$ A=1 php -r 'print_r($_ENV["A"] ?? 0);' // 0
$ A=1 php -r 'print_r(getenv("A"));' // 1
```

nicolas-grekas:
That's because of your ini settings. Still, phpunit populates it. See just below your link above:
https://github.com/sebastianbergmann/phpunit/blob/4f44ade4cf58e1fcddf68b3db8467fe1694552ea/src/Util/Configuration.php#L595

maidmaid:
My bad :

```
$ A=1 php -dvariables_order=E -r 'print_r($_ENV["A"]);' // 1
```

So, if I understand properly, we should use $_ENV in this PR (and expect that users define `-dvariables_order=E` in their settings)?

nicolas-grekas:
Please read the code linked, just below the target line.

stof:
@nicolas-grekas `$_ENV` will indeed work with PHPUnit. My concern is whether it will actually work with **actual** env variables (rather than config-based ones), without requiring to have `E` in the variable orders (which may not be the case by default on all systems)

maidmaid:
>```xml
><php>
>  <env name="foo" value="bar"/>
></php>
>```
>
>The XML configuration above corresponds to the following PHP code:
>
>```php
>$_ENV['foo'] = 'bar';
>```

https://phpunit.readthedocs.io/en/7.3/configuration.html?highlight=env#setting-php-ini-settings-constants-and-global-variables

So, using $_SERVER doesn't work with the env vars defined in the XML configuration, right ?

dunglas:
Closing as it doesn't seem to be a problem anymore. Feel free to reopen if I'm wrong.

PR #113, opened 2018-10-03
Pause on fail

maidmaid:
Currently, it's a little bit boring when a failure/error occurred because there is no way for investigating it at the moment through the browser because it is killed. This PR add an interactive mode by making a pause after every failure/error.

```
➜ PANTHER_NO_HEADLESS=1 vendor/bin/phpunit --debug

Test 'App\AdminTest::testAdmin' started
Error: ko

Press enter to continue...
```

This interactive mode works when the web server is started, the debug option used and the no headless env var set.

### DIFF
--- a/src/ServerTrait.php
+++ b/src/ServerTrait.php
@@ -20,6 +20,8 @@ namespace Symfony\Component\Panther;
  */
 trait ServerTrait
 {
+    public $testing = false;
+
     private function keepServerOnTeardown(): void
     {
         PantherTestCase::$stopServerOnTeardown = false;
@@ -28,5 +30,18 @@ trait ServerTrait
     private function stopWebServer(): void
     {
         PantherTestCase::stopWebServer();
+    }
+
+    private function pause($message): void
+    {
+        if (PantherTestCase::isWebServerStarted()
+            && \in_array('--debug', $_SERVER['argv'], true)
+            && $_SERVER['PANTHER_NO_HEADLESS'] ?? false
+        ) {
+            echo "$message\n\nPress enter to continue...";
+            if (!$this->testing) {
+                fgets(STDIN);
+            }
+        }
     }
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,6 @@ install:
     fi
 
 script:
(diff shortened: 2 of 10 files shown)


maidmaid:
> PHP Fatal error:  Declaration of Symfony\Component\Panther\ServerListener::addError(PHPUnit\Framework\Test $test, Throwable $t, float $time): void must be compatible with PHPUnit\Framework\TestListener::addError(PHPUnit\Framework\Test $test, Exception $e, $time) in /home/travis/build/symfony/panther/src/ServerListener.php on line 22

```php
// PHPUnit 6
public function addError(Test $test, \Exception $e, $time);

// PHPUnit 7
public function addError(Test $test, \Throwable $t, float $time): void;
```

How manage that ?

dunglas:
I propose to just drop support for PHPUnit 7. It will not be maintained anymore in February anyway.

src/ServerTrait.php
```diff
@@ -29,4 +29,15 @@ private function stopWebServer(): void
...
     }
+
+    private function pause($message): void
+    {
+        if (PantherTestCase::isWebServerStarted()
+            && \in_array('--debug', $_SERVER['argv'], true)
+            && $_SERVER['PANTHER_NO_HEADLESS'] ?? false
+        ) {
+            echo "$message\n\nPress enter to continue...";
+            fgets(STDIN);
```
dunglas: You could create an `@internal` 'fgetsMock` property, and call it if defined instead of `fgets` or something similar.

maidmaid: I'm sorry, I don't really see what you mean...

maidmaid: ServerTrait is already internal. Could we use a public `testing` property to make simpler the code ?

dunglas: LGTM

maidmaid: Done.

maidmaid:
Tests added.

maidmaid:
Note in the readme pushed :)

dunglas:
Thanks @maidmaid!

PR #116, opened 2018-10-06
Open DevTools when !headless

maidmaid:
`--auto-open-devtools-for-tabs`

> This flag makes Chrome auto-open DevTools window for each tab. It is intended to be used by developers and automation to not require user interaction for opening DevTools.

https://peter.sh/experiments/chromium-command-line-switches/#auto-open-devtools-for-tabs


dunglas:
Good idea!

dunglas:
Thanks @maidmaid!

PR #117, opened 2018-10-06
Add support of keyboard and mouse to client

maidmaid:


maidmaid:
Unfortunately, using the mouse/keyboard API is still boring :

```php
$client->getMouse()->mouseMove(
    $client->getCrawler()->filter('#my-element')->getElement(0)->getCoordinates()
);
```

Why not create some helpful methods (like `waitFor()`), like :

```php
$client->mouseMoveTo($cssSelector = '#my-element');
```

dunglas:
@maidmaid can you add a test for this new interface? And 👍 on my side for the helper methods you suggest.

dunglas:
Thanks @maidmaid

Issue #54, opened 2018-07-11
Missing execution bit from chromedriver preventing its execution


XavierVallot:
Hi,
I'm trying to use Panthère in Symfony on Debian, but I have this error when launching my tests :
`RuntimeException: sh: 1: exec: /srv/app/vendor/symfony/panthere/src/ProcessManager/../../chromedriver-bin/chromedriver_linux64: Permission denied`

Indeed, here's the permission status :
![image](https://user-images.githubusercontent.com/9256415/42582311-14c45cae-852f-11e8-9a44-6422190dbdbb.png)

Is there a way to install chromedriver without setting permissions by hand ?

dunglas:
Is the project in a shared folder? The mount driver you use most likely don't preserve permissions.

Kocal:
Hey, me and @XavierVallot are working together and I just encountered the same issue.
So yes, we are on a shared FS (Vagrant), but I think it's more a composer issue than something else.

When running composer:
![selection_779](https://user-images.githubusercontent.com/2103975/43848051-63d8db30-9b32-11e8-8024-d531c7c2abb7.png)

When manually cloning the repo:
![selection_780](https://user-images.githubusercontent.com/2103975/43848054-6649305e-9b32-11e8-93ab-3ef499aab59a.png)

Inside the VM, we are using php 7.2.7 and composer 1.6.5.

Kocal:
Well...

I just ran composer outside the VM and everything is fine :thinking: 
![selection_782](https://user-images.githubusercontent.com/2103975/43848613-9f9c2dba-9b33-11e8-87ce-dd020e2bbbd8.png)

Outside the VM, I am using php 7.2.8 and composer 1.7.1.

I will try to update VM's composer to the same version, and see what happens.

Kocal:
@dunglas you are right, this problem only happens when we require a dependency inside this shared folder. I tried with non-shared folder `/srv/app2` and it works.

I tried to update our Vagrantfile with this but it does not work:
```diff
config.vm.synced_folder '.', '/srv/app',
  type: 'nfs',
-  mount_options: ['nolock', 'actimeo=1', 'fsc']
+  mount_options: ['nolock', 'actimeo=1', 'fsc', 'exec']
```

Do you have an idea @tristanbes? :thinking:

tristanbes:
maybe @nervo could enlight us on this (seems to happens only on Linux since on my OSX host, I can't reproduce)

Kocal:
[...]

PR #120, opened 2018-10-09
Added a warning about the `unzip` command

Deuchnord:
This pull request adds a warning to the README about the `unzip` command that must be installed on the computer, especially on Linux systems, as proposed in issue #54.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,8 @@ Use [Composer](https://getcomposer.org/) to install Panther in your project. You
     
     composer req --dev symfony/panther
 
-**Warning: make sure you have `unzip` installed on your computer. If you don't, you may have strange issues due to a PHP's zip extension not preserving the permissions of binaries.**
+**Warning:** The `unzip` command must be installed or you will encounter an error similar to `RuntimeException: sh: 1: exec: /app/vendor/symfony/panther/src/ProcessManager/../../chromedriver-bin/chromedriver_linux64: Permission denied` (or `chromedriver_linux64: not found).
+The underlying reason is that PHP's `ZipArchive` doesn't preserve UNIX executable permissions.
 
 ## Basic Usage
 
--- a/README.md
+++ b/README.md
@@ -30,6 +30,8 @@ Use [Composer](https://getcomposer.org/) to install Panther in your project. You
     
     composer req --dev symfony/panther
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -30,6 +30,8 @@ Use [Composer](https://getcomposer.org/) to install Panther in your project. You
     
     composer req --dev symfony/panther
 
+**Warning: make sure you have `unzip` installed on your computer. If you don't, you may have strange issues due to a PHP's zip extension not preserving the permissions of binaries.**
```
dunglas: Can you reword like this:

    **Warning:** The `unzip` command must be installed or you will encounter an error similar to `RuntimeException: sh: 1: exec: /app/vendor/symfony/panther/src/ProcessManager/../../chromedriver-bin/chromedriver_linux64: Permission denied` (or `chromedriver_linux64: not found ).
    The underlying reason is that PHP's `ZipArchive` doesn't preserve exec permissionS.

It will make the solution easier to find when using Google :)

Deuchnord: Sure

Kocal:
Thanks :slightly_smiling_face:

stof:
I would add something about the fact that this concerns \*nix systems only, not Windows. Telling Windows users to install unzip might scare them, as it is easier to use PHP's ZipArchive there (and Composer even uses ZipArchive by default on Windows).
But Windows users won't have issues when using ZipArchive, as Windows does not care about the Unix permissions of the file.

Deuchnord:
@stof you're right, I'm fixing this

README.md
```diff
@@ -30,6 +30,9 @@ Use [Composer](https://getcomposer.org/) to install Panther in your project. You
     
     composer req --dev symfony/panther
 
+**Warning:** On \*nix systems, the `unzip` command must be installed or you will encounter an error similar to `RuntimeException: sh: 1: exec: /app/vendor/symfony/panther/src/ProcessManager/../../chromedriver-bin/chromedriver_linux64: Permission denied` (or `chromedriver_linux64: not found).
```
stof: missing closing backtick at the end of the error message

Deuchnord: Sorry for this

dunglas:
Thanks @Deuchnord!

PR #122, opened 2018-10-10
Fix E2eTest example to use `html()` method

zrashwani:
In `E2eTest` Example, `text()` method was used with `title` - which is hidden text - so the assertion below will always fail:
`$this->assertContains('My Title', $crawler->filter('title')->text());`
as `text()` method will return empty string, the statement should be updated to use `html()` method as mentioned in **Hidden Text** section, to be like:
`$this->assertContains('My Title', $crawler->filter('title')->html());`


dunglas:
Good catch! Thanks @zrashwani

PR #123, opened 2018-10-16
Force adding form values when submitting a Panther form with values

Pierstoval:
First, `$values` is not used when submitting a Panther form, and second, submitting with values is not compatible with WebDriver.

I thought it would be nice to force setting the values like this when submitting the form.

WDYT?


Pierstoval:
Added a test 👍

dunglas:
Thanks @Pierstoval!

PR #124, opened 2018-10-20
Pass the port number to ChromeDriver

crazytoad:
This allows you to specify a different port number that the `ChromeDriver` should listen to and that the `ChromeManager` should try connecting to if you know the default 9515 isn't available or if you want to run multiple instances on one machine.

Before this patch, if you passed a port number to `ChromeManager` it would never successfully connect to the `ChromeDriver` since the driver was always listening to the default port 9515.


dunglas:
Thanks @crazytoad

PR #129, opened 2018-10-29
add php 7.3 to travis

andreybolonin:


Pierstoval:
@andreybolonin Would you mind to fix Travis' integration with PHP 7.3?

.travis.yml
```diff
@@ -9,6 +9,7 @@ jobs:
       env: lint=1
     - php: '7.2'
       env: deps=low
+    - php: '7.3'
```
pluseg: My advice is to add new env var here, for php7.3:
```
- php: '7.3'
  env: xdebug=0
```
It looks dirty, but Travis has some outdated image of php7.3 (RC3 still instead of stable release) and they follow strange policy by including/excluding xdebug in different php-images. In 7.3 there is no xdebug, so no need to remove xdebug.ini below.
```
before_script:
  - if [[ $xdebug != '0' ]]; then phpenv config-rm xdebug.ini; fi
```

pluseg: Even more clear solution:
```
- phpenv config-rm xdebug.ini || echo "xdebug not available"
```

dunglas:
Thanks @andreybolonin

PR #132, opened 2018-11-05
Add support of a REPL mode

maidmaid:
screencast from 05-11-18 14_58_51


maidmaid:
Hi @bobthecow, could you give me your opinion about this PR please ? :)

stof:
this is not a sandbox mode. This is a REPL mode. There is no sandboxing involved.

panther
```diff
@@ -0,0 +1,53 @@
...
+ * file that was distributed with this source code.
+ */
+
+use Psy\Shell;
+use Symfony\Component\Panther\Client;
+use Symfony\Component\Panther\PantherTestCaseTrait;
+
+require file_exists($autoload = __DIR__.'/vendor/autoload.php') ? $autoload : __DIR__.'/../../autoload.php';
+
+if (file_exists($psyshFile = '/usr/local/bin/psysh')) {
```
dunglas: Should be configurable isn’t it?

maidmaid: It's the default path [suggested by PsySH](https://github.com/bobthecow/psysh/wiki/Installation#-installation).
If you really want a configuration for it, how does it look like ? An env var like `PANTHER_PSYSH_BIN` ?

dunglas: Yes and env var would be nice.

maidmaid: Done.

panther
```diff
@@ -0,0 +1,53 @@
...
+ * file that was distributed with this source code.
+ */
+
+use Psy\Shell;
+use Symfony\Component\Panther\Client;
+use Symfony\Component\Panther\PantherTestCaseTrait;
+
+require file_exists($autoload = __DIR__.'/vendor/autoload.php') ? $autoload : __DIR__.'/../../autoload.php';
+
+if ($psyshBin = $_SERVER['PANTHER_PSYSH_BIN'] ?? false) {
```
dunglas: Oh sorry I may have expressed myself poorly: it should still default to the recommended path:

```suggestion
$psyshPath = $_SERVER['PANTHER_PSYSH_BIN'] ?? '/usr/local/bin/psysh';
if (file_exists($$psyshPath)) {
    require $psyshBin;
```

panther
```diff
@@ -0,0 +1,53 @@
...
+    use PantherTestCaseTrait;
+
+    /**
+     * @var Client
+     */
+    public $p;
+
+    public function __construct()
+    {
+        $_SERVER['PANTHER_NO_HEADLESS'] = 1;
```
dunglas:
```suggestion
        $_SERVER['PANTHER_NO_HEADLESS'] = $_SERVER['PANTHER_NO_HEADLESS'] ?? 1;
```

Just in case someone has a valid use case for disabling headless mode (i.e: the box has no X servers).

maidmaid: Done.

README.md
```diff
@@ -130,6 +130,18 @@ class E2eTest extends PantherTestCase
 }
 ```
 
+### REPL Debugger
+
+Panther provides a REPL (Read-Eval-Print Loop) for helping you to debug your end-to-end tests thanks to the `bin/panther` binary.
```
dunglas:
```suggestion
Panther provides a REPL (Read-Eval-Print Loop) for helping you to debug your end-to-end tests thanks to the `bin/panther` script.
```

Nitpicking, but it's a text file :)

maidmaid:
I added a note in the readme. Let me know if you'd like changes.

README.md
```diff
@@ -130,6 +130,18 @@ class E2eTest extends PantherTestCase
 }
 ```
 
+### REPL Debugger
+
```
dunglas: I would even insert the GIF you provided in the PR here.

README.md
```diff
@@ -130,6 +130,18 @@ class E2eTest extends PantherTestCase
...
+Panther provides a REPL (Read-Eval-Print Loop) for helping you to debug your end-to-end tests thanks to the `bin/panther` binary.
+The Panther client is accessible through a `$p` variable. For instance, try to type something like:
```
dunglas: Maybe should we mention to run `composer require psy/psysh` before running the script?

README.md
maidmaid: Done.

panther
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+use Psy\Shell;
+use Symfony\Component\Panther\Client;
+use Symfony\Component\Panther\PantherTestCaseTrait;
+
+require file_exists($autoload = __DIR__.'/vendor/autoload.php') ? $autoload : __DIR__.'/../../autoload.php';
+
+if (file_exists($psyshBin = $_SERVER['PANTHER_PSYSH_BIN'] ?? '/usr/local/bin/psysh')) {
+    require $psyshBin;
```
bobthecow: requiring the phar file, especially by default (regardless of whether there's a local composer dependency installed) feels bad. it's just asking for version conflicts and incompatible third-party libraries. 

your whole session will use whatever version of Symfony components are shipped with the phar you happen to have on your system, rather than the ones that are depended on by your `composer.json`. or, if the local ones win, psysh won't have Symfony components that it expects. depending on how far the local phar version skews from what is in your composer lockfile, they might not be compatible at all, and we really have no way of knowing.

i think i would prefer always requiring a composer-installed `psy/psysh` dependency, rather than using the phar at all.

bobthecow: if we _do_ support requiring the phar directly, we should respect the user's `$PATH` and load it that way. they might have a global composer install, a copy in `/usr/bin`, a per-user bin path in `~/bin`, etc. there's no way to tell, but it's almost certain we'll get it wrong for some users if we try to guess.

maidmaid:
```php
if (0 <= version_compare(Psy\Shell::version, 'v.0.9.10')) {
    echo 'Please upgrade your version of PsySH script.';
    die();
}
```
Will we need to do something like it ?

maidmaid: >  it's almost certain we'll get it wrong for some users

To check it like below is probably safer:
```php
$psyshBin = strtok(exec(sprintf('%s psysh', '\\' === \DIRECTORY_SEPARATOR ? 'where' : 'command -v')), PHP_EOL)
```

panther
```diff
@@ -0,0 +1,58 @@
...
+use Symfony\Component\Panther\PantherTestCaseTrait;
+
+require file_exists($autoload = __DIR__.'/vendor/autoload.php') ? $autoload : __DIR__.'/../../autoload.php';
+
+if (!class_exists(Shell::class)) {
+    $psyshBin = $_SERVER['PANTHER_PSYSH_BIN'] ?? strtok(exec(sprintf('%s psysh', '\\' === \DIRECTORY_SEPARATOR ? 'where' : 'command -v')), PHP_EOL);
+    if (file_exists($psyshBin)) {
+        require $psyshBin;
+        if (version_compare(Shell::VERSION, '0.9.9', '<=')) {
+            fwrite(STDERR, sprintf('Please upgrade the version of your PsySH script at "%s".', $psyshBin).PHP_EOL);
```
dunglas: What about this:

```suggestion
            @trigger_error(sprintf('Please upgrade the version of your PsySH script at "%s" to version 0.9.9 or superior.', $psyshBin), E_USER_DEPRECATED);
```

It will allow the phpunit bridge and monitoring systems to catch the problem.

panther
bobthecow: even if you require that version of psysh, you should still defer to a locally installed composer dependency rather than loading _anything_ from the phar.

bobthecow: also, the scoped phar is a backwards compatibility break, so it will be at least v0.10.0.

panther
dunglas: Could you change this if appropriate (or resolve the discussion if not @maidmaid)? I would like to merge this one and tag a new release before SymfonyCon.

maidmaid: I just did it :)

theofidry:
Why not shipping the scoped PsySH PHAR instead of recommending the users to install the PsySH package?

dunglas:
What do you mean by "shipping"? Copying it in the source tree will be a pain to maintain.

theofidry:
why would it be more annoying than requiring the dev package? Unless @bobthecow provides a shim package as well, it's likely gonna result in conflicts

dunglas:
It's possible to tell composer to download a phar?

theofidry:
That would be a shim package I'm talking about. Alternatively one can use phive or the phive-plugin. I would love to see the phive plugin go another level and allow downloading PHARs directly but we are not there yet

dunglas:
ping @maidmaid

maidmaid:
I think we should give a chance to the scoped PsySH phar (v0.10) and if it is not satisfying enough, we could always rollback and promote PsySH only as a dev dependency when the v1 of Panther will be published.
So let me bump the version to 0.10 and resolve the conflict, and then you could merge this PR.

dunglas:
Any news regarding this PR @maidmaid? Can we merge it?

Issue #118, opened 2018-10-08
WebServerReadinessProbeTrait::waitUntilReady blocks until forever


cedricziel:
If the script that's fetched doesnt execute without an error, the readiness probe loops endlessly.

I'm currently not sure what happens, but i think it might be a issue with my php version.

In my case, the kernel fails completely and a 5xx is returned.

dunglas:
Hi @cedricziel,

Thanks for the report. Could you provide a repository or a script to reproduce?

flyslick:
Hi,

I got the same issue but i tried to change php version and nothing changed.

I don't know if that can help but i got the same issue while trying to make panther work with Docker but everything is working fine without it.

PR #135, opened 2018-11-10
Fix issue #118

GWillmann:
When running the server on Docker, `@file_get_contents($url, false, $context)` in `WebServerReadinessProbeTrait.php` always returns false because of the timeout in `$context` which is too low.
Because of that, the method `waitUntilReady` is stuck in an endless while loop.

Increasing the timeout from 1 to 5 seems to solve the problem.


dunglas:
Thanks @GWillmann

PR #136, opened 2018-11-14
Make it easy to use an external web server

### LINKED ISSUE #114: Secure web server [question]
Is it possible to use a secure web server for the panther client?

I can see that the scheme is hardcoded here: https://github.com/symfony/panther/blob/master/src/PantherTestCaseTrait.php#L114

The reason for the request is that I'm using dunglas/DunglasAngularCsrfBundle with the secure cookie option. I noticed while trying to run tests the cookie was not set when using the http:// scheme.

dunglas:
Add a new option to not start the built-in PHP web server.
Closes #114.


maidmaid:
Thanks @dunglas.

PR #138, opened 2018-11-14
Fix --window-size Chrome argument

dunglas:


src/ProcessManager/ChromeManager.php
```diff
@@ -83,7 +83,7 @@ private function findChromeDriverBinary(): string
     private function getDefaultArguments(): array
     {
         // Enable the headless mode unless PANTHER_NO_HEADLESS is defined
-        $args = ($_SERVER['PANTHER_NO_HEADLESS'] ?? false) ? ['--auto-open-devtools-for-tabs'] : ['--headless', 'window-size=1200,1100', '--disable-gpu'];
+        $args = ($_SERVER['PANTHER_NO_HEADLESS'] ?? false) ? ['--auto-open-devtools-for-tabs'] : ['--headless', '--window-size=1200,1100', '--disable-gpu'];
```
maidmaid: We navigate differently on a responsive website according to the size window... It could be fine to keep the same window size between headless and no headless mode.

dunglas: Agreed. Could you open a PR?

PR #139, opened 2018-11-14
feat: Adding docblocks to ChromeManager class

shreypuranik:
Adding some documentation to ChromeManager class, plus minor formatting tweaks to aid with readability.

### DIFF
--- a/src/ProcessManager/ChromeManager.php
+++ b/src/ProcessManager/ChromeManager.php
@@ -100,7 +100,7 @@ final class ChromeManager implements BrowserManagerInterface
     }
 
     /**
-     * @return Array
+     * @return array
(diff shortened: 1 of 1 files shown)


dunglas:
Hi @shreypuranik, and thanks for your contribution.
However, for this repository, we follow [the Symfony Coding Standards](https://symfony.com/doc/current/contributing/code/standards.html) (modernized, because this repository is for PHP 7.1+, while Symfony Core still support PHP 5).

CS are automatically checked and fixed using our PHP CS Fixer config: https://github.com/symfony/panther/blob/master/.php_cs.dist

The changes you introduce don't respect these CS, it's why Travis is red.

PR #142, opened 2018-11-26
Add Symfony 3.4 support

navitronic:
After coming across the initial work from @emodric in #82, fixing #79, I decided to dig into what the issues were with supporting Symfony 3.4 in this library and came across a couple of issues.

1. When installing dependencies in travis, the prefer-lowest setting in composer was installing `2.x` versions of   certain Symfony libraries for somewhat mysterious reasons (potentially because of fabpot/goutte), causing missing class errors. Eg: https://travis-ci.org/symfony/panther/jobs/428138576#L685

   This problem seemed to be solved by changing composer.json to require `symfony/symfony:^3.4` instead of the   framework bundle.

   I'm not sure if this is the best fix to the solution, so am happy to hear any other ideas/suggestions on how to fix.
   
2. The declaration of the `request` method in `src/Client.php` was incompatible with the lack of type hinting in v3.4  of `Symfony\Component\BrowserKit\Client`. I approached solving this by renaming the existing Panther Client class and optionally defining two separate classes depending on the major version of Symfony in use.
   
   Not sure if this there is a better way to approach handling changes in method declarations like this, but happy to hear any other solutions or suggestions.
   
Keen to hear any feedback on this and thanks for providing a useful library.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
   "require": {
     "php": ">=7.1",
     "facebook/webdriver": "^1.5",
-    "symfony/browser-kit": "^4.0",
+    "symfony/browser-kit": "^3.4 || ^4.0",
     "symfony/polyfill-php72": "^1.9",
     "symfony/process": "^3.4 || ^4.0"
   },
--- a/composer.json
+++ b/composer.json
@@ -45,6 +45,6 @@
     "fabpot/goutte": "^3.2.3",
     "phpunit/phpunit": "^7.3",
     "symfony/css-selector": "^3.4 || ^4.0",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -45,6 +45,6 @@
     "fabpot/goutte": "^3.2.3",
     "phpunit/phpunit": "^7.3",
     "symfony/css-selector": "^3.4 || ^4.0",
-    "symfony/framework-bundle": "^3.4 || ^4.0"
+    "symfony/symfony": "^3.4 || ^4.0"
```
dunglas: Depending of `symfony/symfony` is considered a bad practice for libraries (since a very long time), this cannot be merged as is.

PR #143, opened 2018-11-28
Fix web server health check issue

Pierstoval:
I encountered an issue like `Call to a member function isStarted() on null` when using the `ServerListener` on CI, I think this check will be enough to fix it


dunglas:
Thanks @Pierstoval

PR #144, opened 2018-11-29
Fix invalid XPath expression

navitronic:
This PR fixes an issue with an extra `=` in the XPath expression for retrieving all form fields.

```
Facebook\WebDriver\Exception\InvalidSelectorException: invalid selector: Unable to locate an element with the xpath expression .//input[@name] | .//textarea[@name] | .//select[@name=] | .//button[@name] because of the following error:
SyntaxError: Failed to execute 'evaluate' on 'Document': The string './/input[@name] | .//textarea[@name] | .//select[@name=] | .//button[@name]' is not a valid XPath expression.
  (Session info: chrome=70.0.3538.110)
  (Driver info: chromedriver=2.37.544337 (8c0344a12e552148c185f7d5117db1f28d6c9e85),platform=Mac OS X 10.14.1 x86_64)
```

### DIFF
--- a/src/DomCrawler/Form.php
+++ b/src/DomCrawler/Form.php
@@ -252,7 +252,7 @@ final class Form extends BaseForm
      */
     private function getAllElements(): array
     {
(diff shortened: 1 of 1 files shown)


dunglas:
Good catch, thanks!

PR #160, opened 2019-01-13
Update LICENSE

Lablnet:


fancyweb:
The project started in 2018.

Lablnet:
so i think it should be like 2018-2019 as i update commit

Kocal:
Or we can just keep "present"

PR #161, opened 2019-01-14
feat: handle several clients and web servers in trait

fancyweb:
Currently, you cannot use several clients and web servers with the trait. I think it would be a great feature for the next major release of Panther (because BC).

IMHO, the goal would be to have this feature but to still be able to use Panther as currently. In other words, the feature should be almost invisible to existing users.

I have started to work on the trait to discuss about it. It would involve a lot of changes in tests but let's keep that for later.

👍 or 👎

### DIFF
--- a/src/PantherTestCaseTrait.php
+++ b/src/PantherTestCaseTrait.php
@@ -29,32 +29,34 @@ trait PantherTestCaseTrait
     /**
      * @var bool
      */
-    public static $stopServerOnTeardown = true;
+    public static $stopServersOnTeardown = true;
 
     /**
      * @var string|null
      */
-    protected static $webServerDir;
+    protected static $webServerDir = null;
 
     /**
-     * @var WebServerManager|null
+     * @var WebServerManager[]
      */
-    protected static $webServerManager;
+    protected static $webServerManagers = [];
 
     /**
+     * The last started web server base uri.
+     *
      * @var string|null
      */
     protected static $baseUri;
 
     /**
-     * @var GoutteClient|null
+     * @var GoutteClient[]
      */
-    protected static $goutteClient;
+    protected static $goutteClients = [];
 
     /**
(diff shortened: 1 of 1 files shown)


src/PantherTestCaseTrait.php
```diff
@@ -29,32 +29,34 @@
...
      */
-    protected static $webServerManager;
+    protected static $webServerManagers = [];
 
     /**
+     * The last started web server base uri.
+     *
      * @var string|null
      */
     protected static $baseUri;
```
fancyweb: This property is a problem. If people relies on it, we would maybe need a method to get the "baseUri" depending of options. Another way would be to name the client / web server when you create them.

src/PantherTestCaseTrait.php
```diff
@@ -69,61 +71,69 @@
...
 
-        self::$baseUri = sprintf('http://%s:%s', $options['hostname'], $options['port']);
+            self::$webServerManagers[$baseUri] = $webServerManager;
+        }
+
+        return $baseUri;
     }
 
-    public static function isWebServerStarted()
+    public static function isWebServerStarted(): bool
```
fancyweb: This might also be a problem. Maybe it would need an argument to know if a particular web server was started (so a need to name them ?)

dunglas:
Multiple clients will definitely be very useful (for instance to tests apps using [Mercure](https://mercure.rocks). However, I can't see the use case for several web servers. Could you describe why you need this?

fancyweb:
> However, I can't see the use case for several web servers. Could you describe why you need this?

To have several webservers that use different environment variables.

Example : in a test class, I want to test a scenario with and without a behavior that is toggled by an environment variable.

dunglas:
I merged an alternative implementation to retrive several clients: #219 
Starting several web servers isn't supported yet.

PR #165, opened 2019-02-04
Compatible with PHPUnit 8.0

jverdeyen:
The latest version (8.0) of PHPUnit requires a return type void on tear down and setup functions.

https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/TestCase.php#L401

This is just a start to make Panther compatible with 8.0.

### DIFF
--- a/src/PantherTestCaseTrait.php
+++ b/src/PantherTestCaseTrait.php
@@ -67,7 +67,7 @@ trait PantherTestCaseTrait
         'external_base_uri' => null,
     ];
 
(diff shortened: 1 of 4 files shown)


dunglas:
According to the tests, there are more methods to patch. Do you want to fix them too?

jverdeyen:
> According to the tests, there are more methods to patch. Do you want to fix them too?

Yes, will do!

dunglas:
Thanks @jverdeyen

PR #167, opened 2019-02-13
Bump ChromeDriver version

dunglas:


dunglas:
The failure looks related to a ChromeDriver or a PHP-WebDriver bug (not sure). Should I skip the failing part?

Issue #72, opened 2018-07-29
InvalidArgumentException: The "Symfony\Component\Panther\DomCrawler\Crawler::clear" method is not supported when using WebDriver.


ChangePlaces:
I use the `filter` method to target elements, and I want to clear out the product title, and type "test title2":

`$crawler->filter('#product_title')->clear()->sendKeys('test title2');`

But this gives the error: `InvalidArgumentException: The "Symfony\Component\Panther\DomCrawler\Crawler::clear" method is not supported when using WebDriver.`

I instead have to use:
`$crawler->findElement(WebDriverBy::id('product_title'))->clear()->sendKeys('test title2');`

I think this is an inconsistency which need addressing?

dunglas:
Yes it can surely be easily fixed, a pr will be very appreciated!

ChangePlaces:
the pr would be a rewrite and remove the goutte interface as clearly some of it works and some of it doesn't (the incorrect 200 status, this) - red flags for bad use of an interface.

dunglas:
This one should be an easy pick. As stated in the blog post, and in the version number, it's still young... There are things to improve. Removing the browserkit interface makes no sense, it's the whole purpose of the project. Use the PHP WebDriver API or Puppeteer directly in this case.

bastien70:
Any update about this ?

PR #176, opened 2019-03-18
Add clear support

matthieuwerner:
Related to issue #72.

Add basically a "crawler->filter()->clear()" support.


stof:
this is wrong. `Crawler::clear` in the DomCrawler component is **not** the same than `WebDriver::clear`. Repurposing the `clear` method to do something totally unrelated is breaking the rules of OOP.

PR #177, opened 2019-03-21
Added readinessPath to use custom path for server readiness detection

OwlyCode:
Hello,

I have a use case where my app is behind a security firewall that redirects anonymous users to Google's oauth page. In that case Panther fails to detect that the web server is up.

Here's a proposal to add a setting that allows custom path for readiness check. For instance you could do this : 

```php
$client = static::createPantherClient([
    'readinessPath' => '/ping'
]);

$client->getCookieJar()->set(new Cookie('jwt_token', $this->getTestToken()));
$client->request('GET', '/dashboard');
// [...]
```

And that would check `http://127.0.0.1:48008/ping` instead of `http://127.0.0.1:48008/`

### DIFF
--- a/src/PantherTestCaseTrait.php
+++ b/src/PantherTestCaseTrait.php
@@ -65,6 +65,7 @@ trait PantherTestCaseTrait
         'port' => 9080,
         'router' => '',
         'external_base_uri' => null,
+        'readinessPath' => '',
     ];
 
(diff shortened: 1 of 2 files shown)


OwlyCode:
Hello again, is there anything I can do to help integrating these changes ?

dunglas:
Thanks @OwlyCode

PR #178, opened 2019-03-23
Add DomCrawler/Field/FileFormField

robertfausk:
...cause it is not implemented yet.

### DIFF
--- a/src/DomCrawler/Field/FileFormField.php
+++ b/src/DomCrawler/Field/FileFormField.php
@@ -31,7 +31,7 @@ final class FileFormField extends BaseFileFormField
     {
         $tagName = $this->element->getTagName();
         if ('input' !== $tagName) {
-            throw new \LogicException(\sprintf('An FileFormField can only be created from an input tag (%s given).', $tagName));
+            throw new \LogicException(\sprintf('A FileFormField can only be created from an input tag (%s given).', $tagName));
         }
 
         $type = \strtolower($this->element->getAttribute('type'));
--- a/src/DomCrawler/Field/FileFormField.php
+++ b/src/DomCrawler/Field/FileFormField.php
@@ -22,6 +22,45 @@ final class FileFormField extends BaseFileFormField
 {
     use FormFieldTrait;
 
+    /** @var array */
+    protected $value;
+
+    public function getValue()
+    {
+        return $this->value;
+    }
+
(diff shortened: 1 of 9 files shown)


robertfausk:
I'm unsure about test structure atm. 

Option A: I could put it in ```FormTest::testFormFields``` next to tests for ```ChoiceFormField``` and extend ```fixtures/form.html``` by a file upload field.

Option B: I could put in a separate file ```tests/DomCrawler/Field/FileFormFieldTest.php``` and create a new fixture file ```fixtures/file_form_field.html```.

@dunglas Which one do you prefer?

dunglas:
Both options are ok!

robertfausk:
Thx for ultra fast feedback. I chose option B :)

Concerning tests I still need to implement ```FileFormField::setValue``` for ```testFileUploadWithInvalidValue```. I just recognized this right now.

robertfausk:
> Concerning tests I still need to implement `FileFormField::setValue` for `testFileUploadWithInvalidValue`. I just recognized this right now.

Hmm, I could not find a good solution for different behaviors of goutte and panther when trying to set an invalid file path to file upload. Goutte returns an array with additional information and panther can only return a string and is throwing an exception in current implementation.

If this is ok, then I'm also fine and this PR is ready for review. :relaxed: 
Otherwise any help is appreciated. :sos:

robertfausk:
I have reworked FileFormField.

One problem I encountered is when FormField is "recreated" by a form, e.g. when using in
```PHP
$fileFormField = $form['file_upload'];
$fileFormField->setValue($this->getUploadFilePath(self::$uploadFileName));
$this->assertContains(self::$uploadFileName, $form['file_upload']->getValue());
```
In this case there will be a new instance of FileFormField be created instead of using the already created one. The new one differs from first one. This is cause of ```Symfony\Component\DomCrawler\Form``` which handles a ```FormFieldRegistry``` internally and won't create a new instance.
I've implemented a workaround in this PR which resets the value of FileFormField but there is a drawback: The size of the file is incorrect.
I've not a private field $fields in ```Symfony\Component\Panther\DomCrawler\Form``` cause this would have increased PR size too much IMO. But I can provide a PR for this too if desired.

Hope I'm not completely failing atm.

dunglas:
Thanks @robertfausk

PR #179, opened 2019-03-29
Ensure same behavior of ChoiceFormField::getValue as in goutte client

robertfausk:
...cause it didn't return same values and hadn't tests yet.

Only problem I encountered is when multiple values are initially checked in checkbox with multiple values allowed. In this case Goutte returned one value as string instead of multiple values as an array. I'll investigate this one further - maybe it's a bug in Goutte?!? For now I've skipped test for this case.

I'm keen on feedback! :male_detective:


dunglas:
Thanks @robertfausk

PR #180, opened 2019-03-29
Fix setValue for TextareaFormField and text based InputFormField

robertfausk:
... especially when there is already a value given for the field.

This PR did not take non-text types and html5 types into account.
For a complete list of types see: https://www.w3schools.com/tags/att_input_type.asp#midcontentadcontainer

### DIFF
--- a/src/DomCrawler/Field/FormFieldTrait.php
+++ b/src/DomCrawler/Field/FormFieldTrait.php
@@ -57,7 +57,7 @@ trait FormFieldTrait
     private function setTextValue($value): void
     {
         // Ensure to clean field before sending keys.
-        // Unable use $this->element->clear(); cause it triggers a change event on it's own which is unexpected behavior.
+        // Unable to use $this->element->clear(); because it triggers a change event on it's own which is unexpected behavior.
         $existingValueLength = \strlen($this->getValue());
         $deleteKeys = \str_repeat(WebDriverKeys::BACKSPACE.WebDriverKeys::DELETE, $existingValueLength);
         $this->element->sendKeys($deleteKeys.$value);
--- a/src/DomCrawler/Field/FormFieldTrait.php
+++ b/src/DomCrawler/Field/FormFieldTrait.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace Symfony\Component\Panther\DomCrawler\Field;
 
 use Facebook\WebDriver\WebDriverElement;
+use Facebook\WebDriver\WebDriverKeys;
 use Symfony\Component\Panther\ExceptionThrower;
 
 /**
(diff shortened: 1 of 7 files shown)


src/DomCrawler/Field/FormFieldTrait.php
```diff
@@ -48,13 +49,17 @@ public function getValue()
...
-        \is_bool($value) ? $this->element->click() : $this->element->sendKeys($value);
+        return $this->element->getAttribute('disabled') ?? false;
     }
 
-    public function isDisabled()
+    private function setTextValue($value): void
     {
-        return $this->element->getAttribute('disabled') ?? false;
+        // Ensure to clean field before sending keys.
+        // Unable use $this->element->clear(); cause it triggers a change event on it's own which is unexpected behavior.
```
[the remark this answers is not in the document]
dunglas: Could you merge this suggestion please @robertfausk

robertfausk:
Sorry for delay!

dunglas:
Thanks @robertfausk!

src/DomCrawler/Field/InputFormField.php
```diff
@@ -22,6 +22,17 @@ final class InputFormField extends BaseInputFormField
 {
     use FormFieldTrait;
 
+    public function setValue($value)
+    {
+        if (\in_array($this->element->getAttribute('type'), ['text'], true)) {
```
stof: what about other textual types like `email` or `url` ?

tests/fixtures/textarea-form-field.html
```diff
@@ -0,0 +1,29 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>Form with ChoiceFormField</title>
```
stof: Title does not describe them properly

src/DomCrawler/Field/InputFormField.php
robertfausk: > ... 
> This PR did not take non-text types and html5 types into account.
I didn't want to enlarge this PR too much. 
I'll create an issue for more types and create a PR if I find some time.

tests/fixtures/textarea-form-field.html
robertfausk: You are right. I need to create a new PR because this one is closed, am I right?

dunglas: I'll take care of this

dunglas: Fixed by 7a041cb.

stof: but this PR is *breaking* these types (as they were all considered textual before)

PR #181, opened 2019-04-03
Increase a lot (x4) boot sequence

lyrixx:
1. Use symfony/http-client to get the status of webdriver process
2. Do not wait an extra 1 s

The major point is the first one. I paste here the commit details:

With a simple PHPUnit setup and a 'regular' web page I got the following
numbers:

Before:

> Time: 13.39 seconds, Memory: 6.00MB

After:

> Time: 3.27 seconds, Memory: 6.00MB

I Profiler the page with [blackfire](https://blackfire.io/profiles/60e52757-9b36-4e95-b084-ef34fd83108e/graph).
And I noticed `file_get_contents()` was **very slow**.

I first updated the binary (see previous commit) but without success.

Then I tried to reproduce manually what happend. The chromedrive boot
really fast, and is directly available on HTTP. So there is something wrong
with `file_get_contents()`. I used tcpdump to debug it, and the chrome
driver do well its job. The issue is in PHP :(.

Since we have a nice HTTP client Symfony, let's use it !

Here is the tcpdump of the dial between php and the driver

``` 16:05:44.854533 IP 127.0.0.1.59044 > 127.0.0.1.9515: Flags [P.], seq
1:66, ack 1, win 342, options [nop,nop,TS val 3026213891 ecr 3026213891],
length 65 E..u..@.@.............%+...sO......V.i.....
.`\..`\.GET /status HTTP/1.1 Host: 127.0.0.1:9515 Connection: close

16:05:44.854645 IP 127.0.0.1.9515 > 127.0.0.1.59044: Flags [P.], seq 1:237,
ack 66, win 342, options [nop,nop,TS val 3026213891 ecr 3026213891], length
236 E.. .x@.@..]........%+..O..........V.......
.`\..`\.HTTP/1.1 200 OK Content-Length:133 Content-Type:application/json;
charset=utf-8 Connection:close

{"sessionId":"","status":0,"value":{"build":{"version":"alpha"},"os":{"arch":"x86_64","name":"Linux","version":"4.15.0-46-generic"}}}
16:05:44.854651 IP 127.0.0.1.59044 > 127.0.0.1.9515: Flags [.], ack 237,
win 350, options [nop,nop,TS val 3026213891 ecr 3026213891], length 0
E..4..@.@.............%+....O......^.(.....
.`\..`\.

[< HANG A LOT HERE >]

16:05:54.864184 IP 127.0.0.1.59044 > 127.0.0.1.9515: Flags [F.], seq 66,
ack 237, win 350, options [nop,nop,TS val 3026223901 ecr 3026213891],
length 0 E..4..@.@.............%+....O......^.(.....
.`...`\.
```

---

**EDIT**

I initially thought it will increase only the very boot sequence, but actually it speeds up **eveything**.

With 3 tests:

Before:
> Time: 42.06 seconds, Memory: 6.00MB
After:
> Time: 8.96 seconds, Memory: 6.00MB

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,6 +20,8 @@
     "php": ">=7.1",
     "facebook/webdriver": "^1.5",
     "symfony/browser-kit": "^4.0",
+    "symfony/contracts": "^1.1",
+    "symfony/http-client": "^4.3",
     "symfony/polyfill-php72": "^1.9",
     "symfony/process": "^3.4 || ^4.0"
   },
--- a/src/ProcessManager/WebServerReadinessProbeTrait.php
+++ b/src/ProcessManager/WebServerReadinessProbeTrait.php
@@ -13,7 +13,9 @@ declare(strict_types=1);
 
 namespace Symfony\Component\Panther\ProcessManager;
 
+use Symfony\Component\HttpClient\HttpClient;
 use Symfony\Component\Process\Process;
+use Symfony\Contracts\HttpClient\Exception\ExceptionInterface;
 
 /**
  * @internal
@@ -36,23 +38,45 @@ trait WebServerReadinessProbeTrait
         }
     }
 
-    public function waitUntilReady(Process $process, string $url, bool $ignoreErrors = false): void
+    public function waitUntilReady(Process $process, string $url, bool $ignoreErrors = false, int $timeout = 5): void
     {
-        $context = stream_context_create(['http' => [
-            'ignore_errors' => $ignoreErrors,
-            'protocol_version' => '1.1',
-            'header' => ['Connection: close'],
-            'timeout' => 5,
-        ]]);
+        $client = HttpClient::create(['timeout' => $timeout]);
 
-        while (Process::STATUS_STARTED !== ($status = $process->getStatus()) || false === @file_get_contents($url, false, $context)) {
-            if (Process::STATUS_TERMINATED === $status) {
-                throw new \RuntimeException($process->getErrorOutput(), $process->getExitCode());
+        $start = microtime(true);
+
+        while (true) {
+            $status = $process->getStatus();
+            if (Process::STATUS_STARTED !== $status) {
+                if (microtime(true) - $start >= $timeout) {
+                    throw new \RuntimeException("Chrome could not start (or has crashed) after $timeout seconds.");
+                }
+
+                usleep(1000);
+
+                continue;
             }
 
-            // block until the web server is ready
+            $ready = false;
+
+            $response = $client->request('GET', $url);
+            $e = $statusCode = null;
+            try {
+                if (200 === $statusCode = $response->getStatusCode()) {
+                    return;
+                }
+            } catch (ExceptionInterface $e) {
+            }
+
+            if (microtime(true) - $start >= $timeout) {
+                if ($e) {
+                    $message = $e->getMessage();
+                } else {
+                    $message = "Status code: $statusCode";
+                }
+                throw new \RuntimeException("Could not connect to chrome after $timeout seconds ($message).");
+            }
+
             usleep(1000);
         }
-        sleep(1);
     }
 }
--- a/tests/fixtures/index.php
+++ b/tests/fixtures/index.php
@@ -1,0 +1,14 @@
+<?php
+
+/*
+ * This file is part of the Panther project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+echo 'OK';


composer.json
```diff
@@ -20,6 +20,8 @@
     "php": ">=7.1",
     "facebook/webdriver": "^1.5",
     "symfony/browser-kit": "^4.0",
+    "symfony/contracts": "1.1.x-dev@dev",
+    "symfony/http-client": "4.3.x-dev@dev",
```
lyrixx: YOLO :) I will be better in one month!

dunglas:
Maybe could you use https://symfony.com/blog/new-in-symfony-4-2-wait-until-processes-are-ready too?

stof:
@dunglas this would require dropping support for symfony/process 3.4 LTS. As we already have a working solution here, that would be too bad.

Install version 4.3 of a *new* component in an existing project works without issue. But for existing component, it is harder (projects which are still using `symfony/symfony` would require a upgrading the whole framework for that).

navitronic:
As an aside to the code in the PR, and more about the validity of the changes, I ported the http client changes from this PR to my 3.4 compatible fork and used Guzzle to provide the http client. Happy to report it reduces the time taken to run a suite of 106 tests from `5.96` minutes to `2.84` minutes.

So, thanks a heap for figuring this out 😁

lyrixx:
I removed the commit where I updated the binaries because it breaks the tests:
```
                                                                                                                                                                                              
1) Symfony\Component\Panther\Tests\ClientTest::testCookie with data set #1 (array('Symfony\Component\Panther\Tes...ntTest', 'createPantherClient'), 'Symfony\Component\Panther\Client')       
Facebook\WebDriver\Exception\UnrecognizedExceptionException: invalid argument: invalid 'domain'                                                                                               
  (Session info: headless chrome=73.0.3683.86)                                                                                                                                                
  (Driver info: chromedriver=2.46.628388 (4a34a70827ac54148e092aafb70504c4ea7ae926),platform=Linux 4.15.0-47-generic x86_64)                                                                  
                                                                                                                                                                                              
/home/gregoire/dev/github.com/symfony/panther/vendor/facebook/webdriver/lib/Exception/WebDriverException.php:158                                                                              
/home/gregoire/dev/github.com/symfony/panther/vendor/facebook/webdriver/lib/Remote/HttpCommandExecutor.php:326                                                                                
/home/gregoire/dev/github.com/symfony/panther/vendor/facebook/webdriver/lib/Remote/RemoteWebDriver.php:547                                                                                    
/home/gregoire/dev/github.com/symfony/panther/vendor/facebook/webdriver/lib/Remote/RemoteExecuteMethod.php:40                                                                                 
/home/gregoire/dev/github.com/symfony/panther/vendor/facebook/webdriver/lib/WebDriverOptions.php:55                                                                                           
/home/gregoire/dev/github.com/symfony/panther/src/Cookie/CookieJar.php:39                                                                                                                     
/home/gregoire/dev/github.com/symfony/panther/tests/ClientTest.php:252                                                                                                                        
```

This was not mandatory to make what I wanted ... 

And I also squashed my commits

tests/fixtures/index.php
```diff
@@ -0,0 +1,13 @@
...
+ * This file is part of the Panther project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+echo 'OK';
```
stof: What is this file ?

lyrixx: Just a file that will return a 200. The Tests suite needs it to be sure the test webserver is started

lyrixx: Note: I was not needed before, because the code did not perform any assertion on the status code. Now it does...

dunglas:
Do we merge it now, or do we wait for the release of the HttpClient component?

lyrixx:
What could go wrong if we merge this right now?

dunglas:
@lyrixx we'll have to update `composer.json` when 4.3 will be released, and we'll not be able to tag a release of Panther in the meantime. Not a big issue on my side.

src/ProcessManager/WebServerReadinessProbeTrait.php
```diff
@@ -38,21 +40,41 @@ private function checkPortAvailable(string $hostname, int $port, bool $throw = t
...
-            'ignore_errors' => $ignoreErrors,
-            'protocol_version' => '1.1',
-            'header' => ['Connection: close'],
-            'timeout' => 5,
-        ]]);
-
-        while (Process::STATUS_STARTED !== ($status = $process->getStatus()) || false === @file_get_contents($url, false, $context)) {
-            if (Process::STATUS_TERMINATED === $status) {
-                throw new \RuntimeException($process->getErrorOutput(), $process->getExitCode());
+        $timeout = 5;
```
GaryPEGEOT: Maybe add `$timeout` as a parameter of the method?

lyrixx: fixed, thanks

lyrixx:
BTW, the tests suite use to take [12+ minutes](https://travis-ci.org/symfony/panther/builds). Now it take less [than 5 minutes](https://travis-ci.org/symfony/panther/builds/520863736). Not bad :)

kissifrot:
Is there an ETA for a merge? :)

src/ProcessManager/WebServerReadinessProbeTrait.php
```diff
@@ -36,23 +38,41 @@ private function checkPortAvailable(string $hostname, int $port, bool $throw = t
...
+            } catch (ExceptionInterface $e) {
+            }
+
+            if (microtime(true) - $start >= $timeout) {
+                if ($e) {
+                    $message = $e->getMessage();
+                } else {
+                    $message = "Status code: $statusCode";
+                }
+                throw new \RuntimeException("Could not connect to chrome after 5 seconds ($message).");
```
Lctrs:
```suggestion
                throw new \RuntimeException("Could not connect to chrome after $timeout seconds ($message).");
```

?

dunglas:
@kissifrot `Symfony 4.3 will be a stable version published in May 2019.` :P https://symfony.com/roadmap/4.3

tests/fixtures/index.php
stof: wouldn't this mean that if your homepage has a bug, starting the server does not work ?

dunglas: I agree we should only check that we get an HTTP response, regardless of the status.

kissifrot:
Since symfony is released, can this PR be merged? :pray:

dunglas:
Thanks @lyrixx 🎆

Issue #192, opened 2019-05-09
Using proxy


hectorprats:
Hello, I would like to test how my website works through proxies from different countries. How can I select proxy in chromeclient? I can not find the way.

Loooooouuuuu:
Hi ! I couldn't find anything in the documentation. I found this way : 

```
$client = Client::createChromeClient(null, [
             '--proxy-server=socks://127.0.0.1:9050',
             '--headless',
             "--disable-gpu",
        ]);
```

which seems to work. It would be better if we could manipulate `$capabilities` in ChromeManager.php in the start() function.

hectorprats:
thanks, it works :-)

dunglas:
Would one you be interested in updating the README to explain this? Thanks!

Loooooouuuuu:
@dunglas My english isn't perfect, but I will try !

nouaman1996:
Hi i was wondering if it works for you guys when the proxy need an authentification? i cant make it work, thanks

stof:
@nouaman1996 AFAICT, the chrome binary does not support configuring the proxy credentials through the CLI.

nouaman1996:
@stof  Hi, so any idea how i can use chromeClient or even seleniumClient with a proxy authentification? thanks

PR #194, opened 2019-05-22
Update README.md for passing custom arguments to the chrome binary

Loooooouuuuu:
Updating README.md as @dunglas asked in #192.

This is my first pull request, and English isn't my first language. I hope I did well !

I wanted to add something about having to pass "--headless" and "--disable-gpu" manually but I couldn't find a good way to say it :(

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,17 +215,11 @@ class E2eTest extends PantherTestCase
 }
 ```
 
-### Using a proxy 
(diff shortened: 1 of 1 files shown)


dunglas:
Thanks @Loooooouuuuu. I tweaked your patch to use the dedicated environment variable instead.

PR #198, opened 2019-06-05
Update composer.json

nicolas-grekas:


dunglas:
Thanks @nicolas-grekas!

PR #219, opened 2019-08-21
Mercure and WebSocket testing: allow to create several browser instances

dunglas:
Easily test applications using [Mercure](https://mercure.rocks), WebSocket and similar technologies with Panther!

This PR introduces a new method to create additional extra, isolated, browsers; that can interact with the primary one. For instance, it can be useful to test a chat application:

```php
use Symfony\Component\Panther\PantherTestCase;

class ChatTest extends PantherTestCase
{
    public function testMultiClient(): void
    {

        $client1 = self::createPantherClient()
        $client1->request('GET', '/chat'); 
        // Connect a 2nd user, using an isolated browser
        $crawler2 = self::createAdditionalPantherClient()->request('GET', '/chat');

        // Say hi
        $crawler2->submitForm('Post message', ['message' => 'Hi folks']);

        // Wait for the message to be propagated
        $client1->waitFor('.message');
        $this->assertSelectorTextContains('.message', 'Hi folks'); // Assertions are always executed in the PRIMARY browser; but you can use the crawler of the other browsers directly if needed
    }
}
```

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -71,6 +71,11 @@ final class Client extends AbstractBrowser implements WebDriver, JavaScriptExecu
         $this->baseUri = $baseUri;
     }
 
+    public function getBrowserManager(): BrowserManagerInterface
+    {
+        return $this->browserManager;
+    }
+
     public function __destruct()
     {
         $this->quit();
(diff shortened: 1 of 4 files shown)


dunglas:
Merged in f5f377fde53510968422449a413c93ef3c188b45. It looks like a GitHub bug occurred, that prevents this PR to be marked as merged.

PR #220, opened 2019-08-22
Update CHANGELOG, README and small code quality improvements

dunglas:

PR #221, opened 2019-08-22
Fix typos

teohhanhui:
Some typos which were missed in https://github.com/symfony/panther/pull/220


dunglas:
Thanks @teohhanhui!

PR #222, opened 2019-08-23
Prevent call of self::getClient when not in static context

robertfausk:
...cause it will break when PantherTestCaseTrait is used in a class
which implements and uses getClient with object context.

This maybe-issue was introduced with:
https://github.com/symfony/panther/commit/b9225b01285b67b857fabb7c9211c473a7687e49

In my specific case I have a method in object context wich will break my test application:
https://github.com/robertfausk/mink-panther-driver/blob/master/src/PantherDriver.php#L62

### DIFF
--- a/src/PantherTestCaseTrait.php
+++ b/src/PantherTestCaseTrait.php
@@ -159,7 +159,10 @@ trait PantherTestCaseTrait
         }
 
         if (\is_callable([self::class, 'getClient'])) {
-            return self::getClient(self::$pantherClient);
+            $rm = new \ReflectionMethod(self::class, 'getClient');
+            if ($rm->isStatic()) {
+                return self::getClient(self::$pantherClient);
+            }
         }
 
         return self::$pantherClient;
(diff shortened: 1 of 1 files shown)


dunglas:
This method is always static isn’t it?

robertfausk:
If a class is using the PantherTestCaseTrait and implementing
```PHP
public function getClient()
```
then this method of the implementing class will be called.

E.g. following will result in an error:
```
public function getClient()
{
    $this->doSomething();
}
```

dunglas:
Ok got it. If you both import `PantherTestCaseTrait` and not import `WebTestAssertionsTrait` breaks because you already redefined `getClient` it breaks.

src/PantherTestCaseTrait.php
```diff
@@ -159,7 +159,10 @@ protected static function createPantherClient(array $options = [], array $kernel
         }
 
         if (\is_callable([self::class, 'getClient'])) {
```
dunglas: Can be merged in a single if: `if (\is_callable([self::class, 'getClient']) && (new \ReflectionMethod(self::class, 'getClient'))->isStatic()) {`

robertfausk: Alright. See next commit.

dunglas:
Thanks @robertfausk

PR #224, opened 2019-08-23
Fix createClient() support

dunglas:
Without this patch, the following error is throw: `A client must be set to make assertions on it. Did you forget to call "Symfony\Component\Panther\PantherTestCase::createClient()"?`.

It's because `getClient` is private, and that the trait is imported in both `WebTestCase` and `PantherTestCase`. A better fix would be to make `getClient` private in Symfony. If done, it will be possible to revert this patch.


robertfausk:
You are welcome. 
Can you please tag a new release 0.5.1?

dunglas:
Yes I’m on it!

PR #230, opened 2019-09-04
Crawlers attr() function can return null

rocramer:
Problem: The declared return type of the `attr()` method in `Crawler` is `string`. 

If an attribute could not be found the `getAttribute()` method in `WebDriverElement` returns `null`. That's why `attr()` is throwing an `TypeError` exception even though it should return null. 

Added that the return type can be null.


dunglas:
Thanks @rocramer

PR #236, opened 2019-09-12
Allow waitFor to wait for any present element

kissifrot:
Current `waitfFor()` method only targets visible elements, which makes impossible to check for hidden inputs presence for example.
This patch allows it to target any element.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -277,7 +277,7 @@ final class Client extends AbstractBrowser implements WebDriver, JavaScriptExecu
             : WebDriverBy::xpath($locator);
 
         $this->wait($timeoutInSecond, $intervalInMillisecond)->until(
-            WebDriverExpectedCondition::visibilityOfElementLocated($by)
+            WebDriverExpectedCondition::presenceOfElementLocated($by)
         );
 
         return $this->crawler = $this->createCrawler();
--- a/tests/ClientTest.php
+++ b/tests/ClientTest.php
@@ -73,6 +73,15 @@ class ClientTest extends TestCase
         ];
     }
 
+    public function testWaitForInvisibleElement(): void
+    {
+        $client = self::createPantherClient();
+        $crawler = $client->request('GET', '/waitfor-invisible.html');
+        $c = $client->waitFor('#hello');
+        $this->assertInstanceOf(Crawler::class, $c);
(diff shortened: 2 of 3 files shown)


webignition:
@dunglas Is there a reason why the current `waitFor()` implementation waits until an element has been rendered as opposed to only waiting until an element has been added to the DOM?

I suppose what I'm asking is: why isn't `waitFor()` currently implemented as the method added by this change?

src/Client.php
```diff
@@ -273,6 +273,21 @@ public function waitFor(string $cssSelector, int $timeoutInSecond = 30, int $int
         return $this->crawler = $this->createCrawler();
     }
 
+    /**
+     * @throws NoSuchElementException
+     * @throws TimeOutException
+     *
+     * @return Crawler
+     */
+    public function waitForAny(string $cssSelector, int $timeoutInSecond = 30, int $intervalInMillisecond = 250)
```
webignition: I find the name `waitForAny` a little confusing, particularly when considered in relation to the already-existing `waitFor` method.

`waitFor()` - waits for an element to be rendered
`waitForAny()` - waits for _any_ element to be rendered (incorrect but that was my first guess)

This method waits for an element to be added to the DOM, suggesting:

`waitForElementAddedToDOM()` - unambiguous and more clear but too verbose
`waitForElementPresence()` - possibly better

kissifrot: That's right, renamed

webignition:
@kissifrot :+1: nice change. You should probably also update the PR title also to reflect the new method name.

dunglas:
I agree with @webignition, we should patch the existing method to check for presence instead of visibility. Could you update the PR please @kissifrot? Thanks!

webignition:
@kissifrot @dunglas Perhaps better if I apply these changes to `waitFor()` in PR #239 to avoid expected conflicts arising from two changes modifying the same method?

kissifrot:
@webignition I don't mind, as long as this changes makes it in a tag :)

dunglas:
I merged #239, feel free to rebase this one or open a new one!

kissifrot:
PR description and contents updated :)

dunglas:
Thanks @kissifrot!

PR #239, opened 2019-09-23
Support xpath expressions in Client::waitFor()

### LINKED ISSUE #238: Update waitFor() to accept either css selector or xpath expression?
At present, `waitFor()` accepts as the first argument a CSS selector. This inherently rules out identifying the element to be waited for using an xpath expression.

We can modify the internal behaviour of this method such that it accepts either a CSS selector or xpath expression without affecting any valid current usage.

An xpath expression will always start with a single forward slash whereas a valid CSS selector will never do so, therefore we can reliably determine the locator type and either call `WebDriverBy::cssSelector()` or `WebDriverBy::xpath()` as needed.

Such a change would "break" things for existing use cases where a CSS selector with a leading forward slash is used, however I'm not sure that counts as valid usage that cannot be broken.

webignition:
Fixes #238 

At present, `Client::waitFor()` accepts as the first argument a CSS selector. This inherently rules out identifying the element to be waited for using an xpath expression.

This change allows the first argument of `Client::waitFor()` to be either a CSS selector or xpath expression. 

An xpath expression always starts with a single forward slash whereas a CSS selector never does. This allows us to differentiate between CSS selectors and xpath expressions and to behave accordingly.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -259,14 +259,14 @@ final class Client extends AbstractBrowser implements WebDriver, JavaScriptExecu
     }
 
     /**
-     * @param string $locator The path to an element to be waited for. Can be a CSS selector or Xpath expression.
-     * @param int $timeoutInSecond
-     * @param int $intervalInMillisecond
-     *
-     * @return Crawler
+     * @param string $locator               The path to an element to be waited for. Can be a CSS selector or Xpath expression.
+     * @param int    $timeoutInSecond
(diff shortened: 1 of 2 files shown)


webignition:
@dunglas Done, I've added a description of the `$locator` argument stating that it can be a CSS selector or XPath expression. The other arguments have no description as I feel that they are self-explanatory.

Issue #241, opened 2019-10-04
Feature - Add Extensions to ChromeDriver


mbeurel:
Hi,
It is not possible to add an ChromeOption extension. 
For exemple it's a interesting feature to add proxy connected.
I had to extend Client and ChromeManager to add it.
Otherwise it's a good bundle.

Thank's :+1:

tienvx:
In my case, I want to set experimental options

PR #243, opened 2019-10-14
Support Chrome extensions

tienvx:
For https://github.com/symfony/panther/issues/241


webignition:
@tienvx You're introducting two changes here, one to allow the setting of extensions and another to allow the setting of experimental options.

Although each change is small, each is also distinct. These two distinct changes may be better introduced in separate PRs - one PR for the setting of extensions, one PR for the setting of experimental options.

src/ProcessManager/ChromeManager.php
```diff
@@ -30,13 +30,26 @@ final class ChromeManager implements BrowserManagerInterface
...
+    private $extensions = [];
+
     public function __construct(?string $chromeDriverBinary = null, ?array $arguments = null, array $options = [])
     {
         $this->options = array_merge($this->getDefaultOptions(), $options);
         $this->process = new Process([$chromeDriverBinary ?: $this->findChromeDriverBinary(), '--port='.$this->options['port']], null, null, null, null);
         $this->arguments = $arguments ?? $this->getDefaultArguments();
     }
 
+    public function setExtensions(array $extensions)
```
webignition: The `$extensions` argument should be an array of strings with each string being a path to an extension.

This need is not clear from the method name alone. Could you document the argument needs?

src/ProcessManager/ChromeManager.php
```diff
@@ -30,13 +30,26 @@ final class ChromeManager implements BrowserManagerInterface
...
+    {
+        $this->extensions = $extensions;
```
webignition: Is there a need to verify that the paths in `$extensions` reference files that can be read? What happens if an invalid or unreadable path is present?

tests/ProcessManager/ChromeManagerTest.php
```diff
@@ -68,4 +68,22 @@ public function testMultipleInstances()
         $driver1->quit();
         $driver2->quit();
     }
+
+    public function testExtensionsAndExperimentalOptions()
```
webignition: This test is examining the behaviour of two methods. Might be best to split the test into two methods where each test examines the behaviour of just one method.

- `testSetExtensions()`
- `testSetExperimentalOptions()`

tests/ProcessManager/ChromeManagerTest.php
```diff
@@ -68,4 +68,22 @@ public function testMultipleInstances()
...
+    {
+        $manager = new ChromeManager();
+        $manager->setExtensions([
+            __DIR__.'/../fixtures/chrome/extension1.crx',
+            __DIR__.'/../fixtures/chrome/extension2.crx',
```
webignition: Are two extensions needed for this test? Would just a single extension work just as well?

tests/ProcessManager/ChromeManagerTest.php
```diff
@@ -68,4 +68,22 @@ public function testMultipleInstances()
...
+        ]);
+        $manager->setExperimentalOptions([
+            'excludeSwitches' => ['enable-automation'],
+            'mobileEmulation' => [
+                'deviceName' => 'Pixel 2',
+            ],
+        ]);
+        $client = $manager->start();
+        $this->assertNotEmpty($client->getCurrentURL());
```
webignition: Is it possible to add further assertions to verify that the expected extensions have been set and that the expected experimental options are in effect?

tienvx: Currently Chrome headless mode does not support extension, so it's not possible to verify that the extension is enabled. We can use `xvfb-run` like [this](https://stackoverflow.com/a/46475980/2622199), but I think it will make thing more complicated.

tienvx:
I split this PR into 2 PRs. The related PR is https://github.com/symfony/panther/pull/276

src/ProcessManager/ChromeManager.php
```diff
@@ -30,13 +30,28 @@ final class ChromeManager implements BrowserManagerInterface
...
     }
 
+    /**
+     * @param string[] $extensions Paths to packed Chrome extensions
+     */
+    public function setExtensions(array $extensions)
+    {
+        foreach ($extensions as $path) {
+            if (!file_exists($path)) {
+                throw new \InvalidArgumentException(\sprintf('The Chrome extension "%s" does not exist.', $path));
```
webignition: Nice changes to the original PR!

Can there be an additional test verifying that this exception is indeed thrown?

Doing so ensures that future changes can't inadvertently break this logic.

tienvx: I added new test case. Thanks for reviewing!

dunglas:
I pushed a commit to improve the public API and make this option usable from a test case, but even in non-headless mode, this doesn't work for me:

Capture d’écran 2020-10-02 à 19 30 01

Could you take a look @tienvx?

Issue #245, opened 2019-10-23
Travis build failures: This version of ChromeDriver only supports Chrome version 76


webignition:
A project of mine using panther very recently started encountering [failing builds on Travis CI](https://travis-ci.com/webignition/php-basil-compilable-source-factory/builds/133174479#L368).

Travis installs the Chrome stable from https://dl.google.com/dl/linux/direct/google-chrome-stable_current_amd64.deb.

Chrome stable recently changed from version 77 to version 78. This seems to be no longer compatible with the version of ChromeDriver supplied by panther.

At the time of writing, https://chromedriver.storage.googleapis.com/LATEST_RELEASE returns `78.0.3904.70` suggesting that there is a newer version of ChromeDriver that can/should be used.

@dunglas What's the best method for updating the version of ChromeDriver supplied with panther? Should I simply run `chromedriver-bin/update.sh` and then create a PR from the resulting changes?

PR #247, opened 2019-10-23
phpstan: ignore unmatched ignored errors

### LINKED ISSUE #246: Builds failing on Travis when linting PHP 7.3
Current PR builds are failing on Travis when linting under PHP 7.3.

The failure is due to phpstan reporting on unmatched ignored error patterns. The ignored error patterns are needed when linting PHP 7.1 and PHP 7.2 but not for PHP 7.3.

phpstan is currently only invoked for PHP 7.3. The solution depends on whether this is to remain the case.

If phpstan is only ever to be invoked for PHP 7.3 we can remove the relevant ignored error patterns. 

If phpstan is to be invoked for PHP < 7.3 we should set `reportUnmatchedIgnoredErrors: false` in `phpstan.neon`.

webignition:
Fixes #246 

phpstan is configured to ignore some errors. Of these, some are only present under PHP 7.1 or 7.2 and are not present under PHP 7.3.

By default phpstan will return a non-zero return code if an ignored error pattern is not matched. This is currently the case when invoking phpstan on travis under PHP 7.3 and is causing all current PR builds to fail.

This change adds `reportUnmatchedIgnoredErrors: false` to the phpstan configuration such that phpstan will be silent about unmatched ignored errors.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -2,6 +2,7 @@ parameters:
 	autoload_files:
 		- vendor/bin/.phpunit/phpunit-7.4/vendor/autoload.php
 	inferPrivatePropertyTypeFromConstructor: true
+	reportUnmatchedIgnoredErrors: false
 	ignoreErrors:
 		# https://github.com/symfony/symfony/pull/33289
 		- '#Parameter \#1 \$value of method Symfony\\Component\\Panther\\DomCrawler\\Field\\FileFormField::setValue\(\) expects string, null given\.#'


webignition:
Build failing currently due to Travis installing Chrome 78 and the version of ChromeDriver supplied with Panther does not appear to support this version of Chrome.

See #245 for details of that issue.

We will need a single PR that addresses both #245 and #246 at the same time to get travis builds passing.

webignition:
Should close this PR if #248 passes.

PR #248, opened 2019-10-23
Bump ChromeDriver to v78, ignore unmatched errors under phpstan

### LINKED ISSUE #245: Travis build failures: This version of ChromeDriver only supports Chrome version 76

### LINKED ISSUE #246: Builds failing on Travis when linting PHP 7.3

webignition:
Fixes #245 
Fixes #246 

This PR addresses two separate issues that are causing/will cause Travis builds to fail. These issues need to be addressed simultaneously to result build failures.

#245 relates to Travis now installing Chrome 78 which requires an update of ChromeDriver. Travis previously installed Chrome 76 then Chrome 77 both of which were compatible with the version of ChromeDriver being used.

#246 relates to phpstan under PHP 7.3 failing due to there being unmatched ignored error patterns (which are required when invoking phpstan under PHP 7.1 or 7.2).

### DIFF
--- a/chromedriver-bin/notes.txt
+++ b/chromedriver-bin/notes.txt
@@ -1,22 +1,12 @@
----------ChromeDriver 76.0.3809.68 (2019-07-16)---------
-Supports Chrome version 76
-Resolved issue 2995: javascript error: circular reference with ChromeDriver 76.0.3809.25 [Pri-1]
-Resolved issue 3007: ChromeDriver 75 - Running Javascript kills driver [Pri-1]
-Resolved issue 2943: goog:chromeOptions.w3c=false doesn't work for POST request with empty body [Pri-1]
-Resolved issue 2947: No W3C compliant endpoints for retrieving logs [Pri-2]
-Resolved issue 1897: Implement Actions API [Pri-1]
-Resolved issue 3011: Perform Actions cannot move pointer to elements in shadow DOM [Pri-2]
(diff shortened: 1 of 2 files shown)


webignition:
@dunglas This addresses two separate issues that are causing all Travis builds to fail and which will resolve Travis build errors for all projects that depend on panther.

dunglas:
Thanks @webignition!

PR #249, opened 2019-10-24
Update composer.json

andreybolonin:
add symfony5 compatibility

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,11 +19,11 @@
   "require": {
     "php": ">=7.1",
     "facebook/webdriver": "^1.7.1",
-    "symfony/browser-kit": "^4.3",
-    "symfony/dom-crawler": "^4.3",
-    "symfony/http-client": "^4.3",
+    "symfony/browser-kit": "^4.3|^5.0",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -19,11 +19,11 @@
   "require": {
     "php": ">=7.1",
     "facebook/webdriver": "^1.7.1",
-    "symfony/browser-kit": "^4.3",
-    "symfony/dom-crawler": "^4.3",
-    "symfony/http-client": "^4.3",
+    "symfony/browser-kit": "^4.3|^5.0",
+    "symfony/dom-crawler": "^4.3|^5.0",
+    "symfony/http-client": "^4.3|^5.0",
```
nicolas-grekas: It would be consistent to use the same style as below: double pipe+spaces around.

manzolo:
To pass tests it is necessary to solve this problem https://github.com/symfony/panther/issues/245

webignition:
PR #248 fixes #245 and  #246 which are both needed to be resolved for this build to pass

dunglas:
Thanks @andreybolonin

andreybolonin:
@dunglas thx

PR #265, opened 2019-11-22
Check if webserver process terminated

pierredup:
When starting the webserver and it terminates, currently you have to wait for the entire timeout before getting any feedback. So this change proposes to throw an exception immediately if the process terminated and get feedback sooner


webignition:
Nice, looks really useful!

Can you add any tests to demonstrate that this change works as expected?

Doing so helps prove that the change is sound and helps prevent future changes breaking this new functionality.

dunglas:
Sorry I forgot about this PR. This has been fixed in https://github.com/symfony/panther/pull/321.

PR #267, opened 2019-12-06
Update .travis.yml

andreybolonin:


.travis.yml
```diff
@@ -10,6 +10,8 @@ jobs:
       env: lint=1
     - php: '7.3'
       env: deps=low
```
dunglas: This line must be removed then.

andreybolonin: done

PR #268, opened 2019-12-11
Throw explanative `LogicException` when driver is not started yet

rvanlaak:
When `executeScript` gets called before the client made a request, an exception should get thrown. The problem; the actual `RuntimeException` can not get created because driver is `null`.

This PR clarifies that these methods can not get called before the driver is started.

Failing test on linting / phpstan does not seem related (?). Is it because of the 0.12 phpstan release of several days ago?

### DIFF
--- a/tests/ClientTest.php
+++ b/tests/ClientTest.php
@@ -17,6 +17,7 @@ use Facebook\WebDriver\Exception\InvalidSelectorException;
 use Facebook\WebDriver\JavaScriptExecutor;
 use Facebook\WebDriver\WebDriver;
 use Facebook\WebDriver\WebDriverExpectedCondition;
+use LogicException;
 use Symfony\Component\BrowserKit\Client as BrowserKitClient;
 use Symfony\Component\BrowserKit\Cookie;
 use Symfony\Component\BrowserKit\CookieJar as BrowserKitCookieJar;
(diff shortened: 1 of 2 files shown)


tests/ClientTest.php
```diff
@@ -17,7 +17,8 @@
 use Facebook\WebDriver\JavaScriptExecutor;
 use Facebook\WebDriver\WebDriver;
 use Facebook\WebDriver\WebDriverExpectedCondition;
-use Symfony\Component\BrowserKit\Client as BrowserKitClient;
+use LogicException;
```
dunglas: We use `\LogicException` instead of this in Symfony.

rvanlaak: As in, you mean not importing the exception class? Ok, will change that.

rvanlaak:
> LGTM, but we need to fix the tests before merging (in another PR).

This PR actually fixes CI for PHP 7.2 , which is failing on `master`. The failing has to do with PHPStan: https://travis-ci.org/symfony/panther/jobs/626071903?utm_medium=notification&utm_source=github_status

@dunglas would you be ok with me adding PHPStan in a separate PR and run it via Github Actions? See https://github.com/php-translation/symfony-storage/pull/38 and https://github.com/theiconic/name-parser/pull/41 and https://github.com/dmishh/SettingsBundle/pull/110 as examples 👍

PR #273, opened 2019-12-17
Run PHPStan via Github Action

rvanlaak:
see https://github.com/symfony/panther/pull/268#issuecomment-566436814

### DIFF
--- a/.github/workflows/static-code-analysis.yaml
+++ b/.github/workflows/static-code-analysis.yaml
@@ -1,0 +1,13 @@
+name: Static code analysis
+
+on: [pull_request]
+
+jobs:
+    static-code-analysis:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@master
+            - name: Run PHPStan
+              uses: docker://jakzal/phpqa:alpine
+              with:
+                  args: phpstan analyze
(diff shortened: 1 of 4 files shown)


.github/workflows/static-code-analysis.yaml
```diff
@@ -0,0 +1,16 @@
+name: Static code analysis
+
+on: [pull_request]
+
+jobs:
+    static-code-analysis:
+        strategy:
+            matrix:
+                php: [7.1, 7.2, 7.3, 7.4]
```
stof: no need to run phpstan 4 times. The result will be the same each time AFAIK.

rvanlaak: I actually did have additional errors while running PHPStan with the PHP7.4 image.

@jakzal did mention changing the image to be running `composer install` in https://github.com/php-translation/symfony-storage/pull/38#discussion_r356660383

Will revert the matrix for now 👍

stof: Were the different images using the same version of PHPStan ?

phpstan.neon
```diff
@@ -1,4 +1,8 @@
 parameters:
+	level: 6
+    paths:
```
stof: indentation looks wrong here

rvanlaak: Thanks. Fixed

dunglas:
Closed by #372. Thanks for your contribution!

PR #276, opened 2019-12-21
Support Chrome experimental options

tienvx:


dunglas:
Could you change this in a similar way to what I did in #243 please?

PR #284, opened 2020-01-17
Rename php-webdriver package

OndraM:
Php-webdriver package [has been renamed](https://github.com/php-webdriver/php-webdriver/issues/730#issuecomment-575601572) and the original is now marked as abandoned.


chalasr:
Thank you @OndraM!

PR #259, opened 2019-11-07
Pass timeout options to RemoteWebDriver

metaer:
Allows to set timeouts like this:
``` php
$client = Client::createChromeClient(null, [], [
    'connection_timeout_in_ms' => 5000,
    'request_timeout_in_ms' => 120000,
]);
```

Partially fixes https://github.com/symfony/panther/issues/155 (one case)

PR #288, opened 2020-01-20
Pass timeouts to RemoteWebDriver (Selenium)

rocramer:
The PR #259 that got merged today introduced the possibility to add timeouts to the RemoteWebDriver. Unfortunately only for the standalone chrome browser. This PR adds the timeout parameter for selenium as well.


.github/workflows/php.yml
```diff
@@ -0,0 +1,30 @@
+name: PHP Composer
```
stof: Changing such CI setup is totally unrelated to passing timeouts to RemoteWebDriver. There is no reason to include that in the same PR IMO.

rocramer: Of course. Just played around with Github action and didn't want to include that in this PR. Will remove that. Sorry for that.

dunglas:
Thanks @rocramer

PR #290, opened 2020-01-22
Pass custom env to webserver

gseidel:
Sometimes the webserver needs to run with custom env vars, than the one used by the current process.

* Allow to pass custom env vars to `WebServerManager`
* Add test to pass `PANTHER_APP_ENV` to Webserver


gseidel:
@dunglas After more work then i thought, I fixed all the deprecation warnings, but the build still failed on some other tests. I think restart the build will help to pass all tests.

dunglas:
Thanks @gseidel

PR #293, opened 2020-02-09
Replace Goutte with HttpBrowser

HypeMC:
It seems that Goutte is being replaced by Symfony's HttpBrowser. This is evident due to the fact that the latest version of Goutte actually only [extends HttpBrowser](https://github.com/FriendsOfPHP/Goutte/blob/v4.0.0/Goutte/Client.php#L19) without adding any additional functionality to it.

This PR deprecates Goutte in favor of Symfony's HttpBrowser.


dunglas:
Awesome, thank you!

PR #303, opened 2020-02-22
Fix an issue when one uses only the PantherTestCaseTrait

Pierstoval:
On a project I just use the trait without the full test case, and if so, the  `self::CHROME` instruction throws an error since this constant is defined in the full test case.


dunglas:
Good catch! Thanks Alex.

PR #305, opened 2020-02-25
Fixed typo

pborreli:


webignition:
Can you change the commit message to something more meaningful?

The commit message of "Fixed typo" won't provide much meaningful context when this is merged.

This charge corrects the panthet web server port number in the "Gitlab CI Integration" section of the readme. 

A commit message that reflects this, perhaps "Correct Gitlab CI Integration panther port number", seems more appropriate.

webignition:
I'm addition, it's generally considered good practice to first open an issue when finding a bug and then subsequently creating a PR to resolve the reported issue, with the PR description referencing the relevant issue (such as "Fixes #<issue-number>).

Doing so allows the issue to be de-coupled from the changes required to address the issue which presents a variety of benefits.

dunglas:
Thanks @pborreli.

@webignition, thanks for helping maintaining the project. For such small changes **not** opening an issue is better IMHO, so we can just close or merge. An issue + a PR is two times more work for the maintainer.

Issue #304, opened 2020-02-24
Docker issues


gabrielspiteri-highlight:
Hi 

I am trying to use the panther module inside a docker container. However when I try to instantiate a ChromeClient I am getting the following error message: **session not created: This version of ChromeDriver only supports Chrome version 80**

It would seem that `apt-get install chromium` installs **version 73** which could be why I am mis-matching with the ChromeDriver. 

Anything I can do to get chrome/chromium version 80 installed?

Thanks

gabrielspiteri-highlight:
Update: managed to install version 80 of chrome using the following in my docker file:

# Install chrome, required for the symfony/panther library
RUN wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb -O /tmp/chrome.deb
RUN dpkg -i /tmp/chrome.deb || apt-get install -yf
RUN rm /tmp/chrome.deb

dunglas:
Thanks for the update! Could you open a PR to update the Dockerfile example we provide?

gabrielspiteri-highlight:
@dunglas sure ... my only concern with the code i used is that say google issues a new release for chrome. My docker command would install the latest version which might lead to another mismatch between driver and chrome. Maybe we should lock it down to the version specific deb file that matches the current driver we use?

gabrielspiteri-highlight:
@dunglas I have opened PR as requested. #306

webignition:
@gabrielspiteri-highlight Regarding the chrome/chrome-driver potential incompatibility problem, this can be addressed by always installing both the latest chrome and the latest chrome-driver.

In #306 a URL to download chrome 80 is used. As you said, this will require updating as new chrome releases are made.

Use https://dl.google.com/dl/linux/direct/google-chrome-stable_current_amd64.deb to get the current release regardless of version number.

To ensure that chrome-driver matches, `cd` to `symfony/panther/chromedriver-bin` and run `update.sh` from that directory. This will retrieve the latest chrome driver. How exactly this is to be achieved in the example docker image I can't be certain but I'm guessing you get easily figure that one out.

Kolvin:
Looks like the PR #306 has gone stale, would be great to get it merged to save folk time in the future :)

I have the following working

```dockerfile
FROM php:7.4-fpm
###
RUN apt-get update && apt-get install -y libappindicator1 fonts-liberation \
[...]

PR #306, opened 2020-02-25
Update README.md docker setup documentation

gabrielspiteri-highlight:
@dunglas I have opted to specify version 80 of the chrome installation in order to be consistent with the driver currently used.


README.md
```diff
@@ -317,7 +317,14 @@ Here is a minimal Docker image that can run Panther:
 ```
 FROM php:latest
 
-RUN apt-get update && apt-get install -y libzip-dev zlib1g-dev chromium && docker-php-ext-install zip
+# Install PHP zip extension, a dependency for symfony/panther library
+RUN docker-php-ext-install zip
+
+# Install chrome, required for the symfony/panther library
+RUN wget http://dl.google.com/linux/chrome/deb/pool/main/g/google-chrome-stable/google-chrome-stable_80.0.3987.122-1_amd64.deb -O /tmp/google-chrome-stable_80.0.3987.122-1_amd64.deb
```
webignition: As per my comment in #304, this URL can be changed to one that retrieves the latest release. `https://dl.google.com/dl/linux/direct/google-chrome-stable_current_amd64.deb`. 

A subsequent call to `symfony/panther/chromedriver-bin/update.sh` (executed from the directory where it resides) will ensure that chrome-driver matches the installed chrome version.

gabrielspiteri-highlight: @webignition I have updated README file as requested I agree that download always the stable and running the update.sh is the right approach. In  this PR I am proposing a slight change to the sh file since basically it relies on running CD command before executing it. Otherwise the downloaded files end up in whatever directory you executed the script from. With my changed the script understand where it is and ensure that the new files end up in the same directory.

OGProgrammer:
Dockerfile doesn't work for me, I had to add the apt-get install line again along with wget but maybe cause I'm working on php:apache docker

OGProgrammer:
Also leaving it like `{path-to-vendor-folder}` will confuse people I think so maybe make that little more clear or something, not sure what to do about that.

dunglas:
Could you revert the changes done to the shell script while #358 isn't ready? I would like to merge the better Docker file.

PR #318, opened 2020-03-17
Fix constant access in Trait

### LINKED ISSUE #317: Constant usage in Trait broken
When I'm not using the `PantherTestCase` base class and am just using the `PantherTestCaseTrait` I receive the following error:

```
Error: Undefined class constant 'CHROME'
```

The problem arises because all constants inside the trait are accessed with `self::`. Using the class name instead should fix the problem.

fetzi:
Fixes #317

### DIFF
--- a/src/PantherTestCaseTrait.php
+++ b/src/PantherTestCaseTrait.php
@@ -162,13 +162,13 @@ trait PantherTestCaseTrait
      */
     protected static function createPantherClient(array $options = [], array $kernelOptions = []): PantherClient
     {
-        $browser = ($options['browser'] ?? self::$defaultOptions['browser'] ?? self::CHROME);
+        $browser = ($options['browser'] ?? self::$defaultOptions['browser'] ?? PantherTestCase::CHROME);
         $callGetClient = \is_callable([self::class, 'getClient']) && (new \ReflectionMethod(self::class, 'getClient'))->isStatic();
         if (null !== self::$pantherClient) {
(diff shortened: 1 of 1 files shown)


src/PantherTestCaseTrait.php
```diff
@@ -162,21 +162,21 @@ public static function isWebServerStarted()
...
-        $browser = ($options['browser'] ?? self::$defaultOptions['browser'] ?? self::CHROME);
+        $browser = ($options['browser'] ?? self::$defaultOptions['browser'] ?? PantherTestCase::CHROME);
         $callGetClient = \is_callable([self::class, 'getClient']) && (new \ReflectionMethod(self::class, 'getClient'))->isStatic();
         if (null !== self::$pantherClient) {
             $browserManager = self::$pantherClient->getBrowserManager();
             if (
-                (self::CHROME === $browser && $browserManager instanceof ChromeManager) ||
-                (self::FIREFOX === $browser && $browserManager instanceof FirefoxManager)
+                (PantherTestCase::CHROME === $browser && $browserManager instanceof ChromeManager) ||
+                (PantherTestCase::FIREFOX === $browser && $browserManager instanceof FirefoxManager)
```
nusje2000, line 171: We could also move the constants to this trait. Since this is the only thing that uses these constants, it would make sense to define them here as well.

dunglas:
Fixed by #320.

PR #320, opened 2020-03-27
CHROME & FIREFOX constant late static binding

symfonyaml:
**Context:**
I would like to be able to set the `const CHROME` constant anywhere in the inheritance while using the `PantherTestCaseTrait`.

**Example:**
```php
class CustomPanther extends PantherDriver
{
    const CHROME = 'chrome';
}

class PantherDriver
{
    use PantherTestCaseTrait;
}
```
**Error**
Fatal error: Undefined class constant 'CHROME'

**Solution:**
Be able to set the CHROME constant from the child classes, using **late static bindings**.

### DIFF
--- a/src/PantherTestCaseTrait.php
+++ b/src/PantherTestCaseTrait.php
@@ -168,7 +168,7 @@ trait PantherTestCaseTrait
             $browserManager = self::$pantherClient->getBrowserManager();
             if (
                 (static::CHROME === $browser && $browserManager instanceof ChromeManager) ||
-                (self::FIREFOX === $browser && $browserManager instanceof FirefoxManager)
+                (static::FIREFOX === $browser && $browserManager instanceof FirefoxManager)
             ) {
                 return $callGetClient ? self::getClient(self::$pantherClient) : self::$pantherClient;
             }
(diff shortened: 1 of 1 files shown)


fetzi:
I think the meaning behind the Trait is to use it in any Test class and I think you shouln't have to define this constants in every project. So maybe the solution from #317 would be better. What do you think?

symfonyaml:
- The purpose of this trait is to be re-used, but also also in other case than TestCase. _e.g.:_
https://github.com/robertfausk/mink-panther-driver/blob/master/src/PantherDriver.php#L41
- Since traits can not have constant, the constant will have to be the the class itself. So this constant is a little bit like writing:  `abstract const CHROME;`. Its mandatory, without default value.
- I need to update FIREFOX constants too...

fetzi:
I know that traits cannot have constants defined, but I thought the purpose of the trait would be to use it almost "anywhere". So I don't think it should be required to always define these constants?

symfonyaml:
@fetzi I understand your concerns and what you mean by defining these constants outside the trait.
I just didn't want to break the design pattern chosen by Kevin @dunglas.

Guikingone:
Hi everyone 👋 

Any news about this PR? 

I'm facing the same issue on https://github.com/Guikingone/panther-extension

dunglas:
Thanks @symfonyaml

PR #321, opened 2020-03-27
Improve error message when the web server crashes

dunglas:
Before:

    RuntimeException: Could not start web server (or it crashed) after 30 seconds.

After:

    RuntimeException: Could not start web server. Exit code: 1 (General error). Error output: Directory tests/Application/public does not exist.


dpfaffenbauer:
Oh boy yes, I just had that issue this week and could only resolve it using the debugger. Thought about doing a PR when I find some time.

Big 👍

PR #322, opened 2020-03-29
Fix tests

dunglas:


composer.json
```diff
@@ -20,7 +20,7 @@
     "php": ">=7.1",
     "php-webdriver/webdriver": "^1.8.1",
     "symfony/browser-kit": "^4.3 || ^5.0",
-    "symfony/dom-crawler": "^4.3 || ^5.0",
+    "symfony/dom-crawler": "^4.3 || ^5.0, !=4.4.6, !=5.0.6",
```
dunglas: See https://github.com/symfony/panther/pull/322

PR #323, opened 2020-03-29
Add docs for GitHub actions, and some other small fixes

dunglas:


README.md
```diff
@@ -333,10 +333,35 @@ RUN apk add --no-cache \
...
+jobs:
+  tests:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install dependencies
+        run: composer install -q --no-ansi --no-interaction --no-scripts --no-suggest --no-progress --prefer-dist
```
stof: is `composer` actually preinstalled ? Or do you need the `setup-php` action before that ?

dunglas: It is preinstalled, you don't need anything: https://github.com/actions/virtual-environments/blob/master/images/linux/Ubuntu1804-README.md

Issue #326, opened 2020-04-01
PHP Webserver only handles certain amount of requests


dpfaffenbauer:
I am not sure what is going on, but the PHP Webserver freezes for me when started by Panther at around 384 requests. That can be any kind of requests, css, js or actual PHP. I'll provide a TestCase and see if that also happens on travis.

dpfaffenbauer:
Interesting, this only happens with PHP 7.4. Works fine with 7.3...

PR #327, opened 2020-04-01
Add Test for #326

dpfaffenbauer:
This basically should prove #326. It happens for also with Github Actions.

If I manually start it and use that with Panther, it works.


dunglas:
A workaround could be to use Symfony CLI (if available) instead of the PHP Webserver (which isn't designed to handle high loads). What do you think @fabpot?

dpfaffenbauer:
that is how I solved it currently to run my suite.

I'd be happy to create a PR if symfony cli is a valid option?

fabpot:
Sure, if you have PHP FPM available, it will definitely help.

dpfaffenbauer:
this might be the reason: https://bugs.php.net/bug.php?id=79514

PR #328, opened 2020-04-01
fix email field with existing value

dpfaffenbauer:
If an input field of type "email" already has a value, the field value is not reset, it's getting appended. this fixes it.

### DIFF
--- a/src/DomCrawler/Field/InputFormField.php
+++ b/src/DomCrawler/Field/InputFormField.php
@@ -24,7 +24,7 @@ final class InputFormField extends BaseInputFormField
 
     public function setValue($value)
     {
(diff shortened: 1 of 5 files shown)


dunglas:
This bug probably also affect other types don't you think? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#%3Cinput%3E_types

dpfaffenbauer:
Very likely. But didn’t come so far yet. But I will come along those types as well in the future and could provide separate PRs

dpfaffenbauer:
FYI: added further tests for number

dunglas:
Thanks @dpfaffenbauer

PR #329, opened 2020-04-06
Using Panther without Symfony and without PHPUnit

symfonyaml:
**As a panther user, I want to create tests without symfony and without PHPUnit installed.**

**Problem**:
When I use the `PantherTestCaseTrait` trait without PHPUnit installed, I have this error:
`PHP Fatal error:  Class 'PHPUnit\Framework\TestCase' not found`

Since PHPUnit is not defined as a dependency of this project, I think we should be able to create an empty test case for this matter

**Suggestion**:
I have updated the `PantherTestCase` to check first if `PHPUnit\Framework\TestCase` is defined ( = PHPUnit installed). Otherwise create a simple abstract `PantherTestCase` class, and let the user to create its own test case methods.

**Use case**:
I am writing a Behat + Panther extension, without PHPUnit.

### DIFF
--- a/src/PantherTestCase.php
+++ b/src/PantherTestCase.php
@@ -13,52 +13,59 @@ declare(strict_types=1);
 
 namespace Symfony\Component\Panther;
 
-use PHPUnit\Framework\TestCase;
-use Symfony\Bundle\FrameworkBundle\Test\ForwardCompatTestTrait;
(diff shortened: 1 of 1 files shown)


dunglas:
I don't see the benefits of doing this. This glue code is dedicated for PHPUnit. It is useless without it. It's already possible to integrate Panther in any other testing framework by using the library directly. Closing for now.

PR #330, opened 2020-04-07
Web server options for environment variables

symfonyaml:
**Goal:**
Making some improvements to be able to use Panther as standalone (without Symfony and without PHPUnit).

**Feature:**
I would like to send custom environment variables to the created web server like:
```php
static::createPantherClient([
    'environmentVariables' => ['CUSTOM_ENV' => 'foo'],
]);
```


ThomasTr:
Helps me to execute different test_cases via custom kernel, would like to see this in next release @dunglas  👍

dunglas:
I don't understand where `environmentVariables` is used?

ThomasTr:
I for myself have different test cases, with which I map different test scenarios, e.g. to mock different external API's, load different configs.

`$this->options['environmentVariables']['TEST_CASE'] = $testCase;`

Then i can load the kernel with the specific environment:

`$kernel = new Kernel($_SERVER['TEST_CASE'], 'config.yaml', $_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG']);
`

arderyp:
I am not sure how this works.  Here is my code:

```php
// in my panther test
$this->client = static::createPantherClient([
    'environmentVariables' => [
        "CUSTOM_PATHER_VAR" => "this is a test",
    ],
]);
```

```php
// in my index.php
echo $_SERVER["CUSTOM_PATHER_VAR"];
echo getenv("CUSTOM_PATHER_VAR");
``

```php
// in my Kernel.php
echo $_SERVER["CUSTOM_PATHER_VAR"];
echo getenv("CUSTOM_PATHER_VAR");
``

I am not seeing my custom var (`"this is a test"`) make it into `index.php` or `Kernel.php`.  Am I missing something?

@ThomasTr suggests it should be available in `$_SERVER`, which makes sense, as that's how `APP_ENV` and other env vars are made available, but I'm not finding that to be the case.

**Comment by arderyp** [comment]
I am not sure how this works.  Here is my code:

```php
// in my panther test
$this->client = static::createPantherClient([
    'environmentVariables' => [
        "CUSTOM_PATHER_VAR" => "this is a test",
    ],
]);
```

```php
// in my index.php
echo $_SERVER["CUSTOM_PATHER_VAR"];
echo getenv("CUSTOM_PATHER_VAR");
``

```php
// in my Kernel.php
echo $_SERVER["CUSTOM_PATHER_VAR"];
echo getenv("CUSTOM_PATHER_VAR");
```

I am not seeing my custom var (`"this is a test"`) make it into `index.php` or `Kernel.php`.  Am I missing something?

@ThomasTr suggests it should be available in `$_SERVER`, which makes sense, as that's how `APP_ENV` and other env vars are made available, but I'm not finding that to be the case.

PR #332, opened 2020-04-22
Symfony CLI Webserver 

dpfaffenbauer:
As discussed here: https://github.com/symfony/panther/pull/327#issuecomment-617720461

This adds support for the Symfony CLI Webserver. As there is no check if `symfony server:start` actually can run, I though we just define via option what web-server to start.

So with this, there is a new Option `webServer` (env var: `PANTHER_WEB_SERVER`) which just points to the Webserver class, that class has to implement the interface `WebServerManagerInterface`. That theoretically also allows custom webserver implementations if needed.


src/PantherTestCaseTrait.php
```diff
@@ -140,11 +142,23 @@ public static function startWebServer(array $options = []): void
             'webServerDir' => self::getWebServerDir($options),
             'hostname' => $options['hostname'] ?? self::$defaultOptions['hostname'],
             'port' => (int) ($options['port'] ?? $_SERVER['PANTHER_WEB_SERVER_PORT'] ?? self::$defaultOptions['port']),
-            'router' => $options['router'] ?? $_SERVER['PANTHER_WEB_SERVER_ROUTER'] ?? self::$defaultOptions['router'],
+            'params' => [
+                'router' => $options['router'] ?? $_SERVER['PANTHER_WEB_SERVER_ROUTER'] ?? self::$defaultOptions['router'],
```
dpfaffenbauer, line 146: since router is only needed for the PHP Webserver, I thought it'd be better to introduce a webserver `params` variable here

tests/fixtures/env.php
```diff
@@ -14,7 +14,7 @@
 require __DIR__.'/security-check.php';
 
 if ('APP_ENV' === $_GET['name'] ?? null): ?>
-    <?=$_ENV['APP_ENV']; ?>
+    <?=$_ENV['APP_ENV'] ?? $_SERVER['APP_ENV']; ?>
```
dpfaffenbauer, line 17: not sure why exactly, but with Symfony CLI Webserver, ENV vars passed to Symfony Process are in $_SERVER not $_ENV, but I think that is right, isn't?

tests/fixtures/security-check.php
```diff
@@ -15,7 +15,6 @@
     isset($_SERVER['HTTP_CLIENT_IP'])
     || isset($_SERVER['HTTP_X_FORWARDED_FOR'])
     || '127.0.0.1' !== $_SERVER['REMOTE_ADDR']
-    || PHP_SAPI !== 'cli-server'
```
dpfaffenbauer, line 18: that doesn't work with symfony-cli anymore, since it fpm and not `cli-server` then

dpfaffenbauer:
@dunglas any feedback on this?

src/ProcessManager/SymfonyWebServerManager.php
```diff
@@ -0,0 +1,66 @@
...
+                    '--no-tls'
+                ]
+            ),
+            $documentRoot,
+            $env,
+            null,
+            null
+        );
+
+        // Symfony Process 3.4 BC: In newer versions env variables always inherit,
```
dunglas, line 60: We only support 4.4 now, so this can be dropped.

src/ProcessManager/PHPWebServerManager.php
```diff
@@ -19,27 +19,14 @@
 /**
  * @author Kévin Dunglas <dunglas@gmail.com>
  */
-final class WebServerManager
+final class PHPWebServerManager extends AbstractProcessWebServer
```
dunglas, line 22:
```suggestion
final class PhpWebServerManager extends AbstractProcessWebServer
```

tests/fixtures/env.php
dunglas: Even just `$_SERVER['APP_ENV']` (without the call to `$_ENV` first should do the trick).

tests/fixtures/security-check.php
dunglas: That's a bit annoying. Does Symfony CLI sets an env var or something like this allowing to detect it? /cc @fabpot

dunglas (changes requested):

src/ProcessManager/AbstractProcessWebServer.php
```diff
@@ -0,0 +1,67 @@
...
+use Symfony\Component\Process\Process;
+
+/**
+ * @author Dominik Pfaffenbauer <dominik@pfaffenbauer.at>
+ */
+abstract class AbstractProcessWebServer implements WebServerManagerInterface
+{
+    use WebServerReadinessProbeTrait;
+
+    protected $hostname;
```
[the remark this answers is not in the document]
dpfaffenbauer: So we use PHP 7.4 as min?

tests/fixtures/security-check.php
dpfaffenbauer: `strpos($_SERVER['SERVER_SOFTWARE'], 'Symfony Local Server')` might work, WDYT?

tests/fixtures/env.php
dpfaffenbauer: did not work....

src/PantherTestCaseTrait.php
dpfaffenbauer: reverted and added --passthru to symfony-cli webserver

src/PantherTestCaseTrait.php
```diff
@@ -140,11 +142,23 @@ public static function startWebServer(array $options = []): void
...
+            ],
             'readinessPath' => $options['readinessPath'] ?? $_SERVER['PANTHER_READINESS_PATH'] ?? self::$defaultOptions['readinessPath'],
         ];
 
-        self::$webServerManager = new WebServerManager(...array_values($options));
+        $webServer = $options['webServer'] ?? $_SERVER['PANTHER_WEB_SERVER'] ?? self::$defaultOptions['webServer'];
+
+        if (!class_exists($webServer)) {
```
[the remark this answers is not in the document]
dpfaffenbauer: created a factory, can you check if that is how you imagined?

dpfaffenbauer:
@dunglas why was this closed without any comment?

dunglas:
Sorry about that. It has been closed automatically by GitHub because we rewrote the history of the `main` branch to remove the binaries (to reduce the size of the repository). But it was useless as `git` also downloads tags by default. Could you rebase and reopen?

dpfaffenbauer:
Alright. Yes will rebase it and open a new PR.

dunglas:
You should be able to reopen this one (to preserve the history). If GutHub blocks it, I can do it.

dpfaffenbauer:
👍  will let you know if I run into problems

dpfaffenbauer:
@dunglas just rebased the branch, but github doesn't seem to pick it up here: https://github.com/symfony/panther/compare/main...dpfaffenbauer:symfony-cli-webserver

PR #343, opened 2020-05-20
Update chromedriver to version 83

christiaan:
Our builds on travis are broken because the version of Chrome on travis is newer than the one included with panther.


christiaan:
@dunglas any chance this can be merged? Or is there a different channel for the chromedriver to be updated?

dunglas:
Thanks @christiaan

christiaan:
@dunglas thinking about this a bit more it seems like a security issue to me.
I could have easily added malicious code to the binary and nothing guards against that. As long as it functions still as a chrome driver.

What is the reason the chromedriver is included in this package?
Would it be a more secure idea to download the chromedriver on first use when no `PANTHER_CHROME_DRIVER_BINARY` is set?  Assuming that the https urls in https://github.com/symfony/panther/blob/master/chromedriver-bin/update.sh provide enough guarantee.

jdreesen:
As it's a shell script, it won't work on Windows, though.

christiaan:
I wasn't suggesting that it should call that script. Because when running we know the environment so it only has to download the driver for the current CPU architecture. Which can be handled directly in PHP no need for a script.

dunglas:
I decided to include the binaries for convenience. But with the rise of Docker and GitHub Actions it's less and less necessary. I agree that we should remove the binaries from the repository, use the `chromedriver` / `geckodriver` available in the path if they exist, and provide a command to download the latest version (a port of the existing shell script in PHP) as a fallback.

Is someone interested to work on this?

PR #346, opened 2020-06-01
Single instruction to update driver

gseidel:
My travis test fail sometimes, because i don't have the latest version of the chrome driver.

``This version of ChromeDriver only supports Chrome version 80``

So i want to execute the update scripts for the binaries before i execute my scripts.

``
./vendor/symfony/panther/chromedriver-bin/update.sh
``

But they will install everything in my project directory. I know i can use ``cd`` the go in the directory first. But i think this line makes it more nice to use this script.

I also just thinking to avoid this kind of error and to prevent executable pollution in this repository, by update this binaries if they don't exists (First use of panther). And if some driver version problem occurs, just provide this script by adding some information about it to the readme. Just let me know what you think. I could create further PR.

### DIFF
--- a/chromedriver-bin/update.sh
+++ b/chromedriver-bin/update.sh
@@ -2,6 +2,8 @@
 # Author: Kévin Dunglas <dunglas@gmail.com>
 # Download the last version of ChromeDriver binaries
 
+cd "$(dirname "$0")"
+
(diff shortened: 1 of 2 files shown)


dunglas:
Thanks @gseidel

PR #353, opened 2020-06-26
allow to customize driver's arguments

Lctrs:
Useful if you want to turn on logging for example.


Lctrs:
@dunglas How would you implement it ? I've never implemented a test with custom env vars.

webignition:
@Lctrs Regarding tests, I have a couple of suggestions.

Firstly, `ChromeManager::getChromeDriverArguments()` is at present hard to test. This is a private method that affects a private property of a private property of `ChromeManager`. No getters are present to examine the state of these private properties.

Try making `ChromeManager::getChromeDriverArguments()` directly testable by factoring it out to a separate class that can be tested and inject that class into the `ChromeManager` via the constructor.

For example, say you created a class named `ChromeDriverArgumentFactory()` with a public method named `create()`. It should be relatively straightforward within a unit test to mutate `$_SERVER['PANTHER_CHROME_DRIVER_ARGUMENTS']` as needed, create an instance of the factory, call the `create()` method and verify that the return value is as expected.

Secondly, what chrome driver arguments are you needing to set? Are you able to describe how you would tell the difference between two `ChromeManager` instances, one with a given driver argument set and one without? When executing panther when a given chrome driver argument is set, how might you determine from the behaviour of chrome that the chosen chrome driver argument was actually set correctly?

src/ProcessManager/ChromeManager.php
```diff
@@ -83,6 +92,19 @@ private function findChromeDriverBinary(): string
         }
     }
 
+    private function getChromeDriverArguments(): array
+    {
+        $args = ['--port='.$this->options['port']];
+
+        // Add custom arguments with PANTHER_CHROME_DRIVER_ARGUMENTS
+        if ($_SERVER['PANTHER_CHROME_DRIVER_ARGUMENTS'] ?? false) {
+            $arguments = explode(' ', $_SERVER['PANTHER_CHROME_DRIVER_ARGUMENTS']);
```
dunglas, line 101: It may not work for complex arguments (values containing a space between quotes for instance). Shouldn't we find something smarter here?

PR #354, opened 2020-06-26
client: throw an exception when getting history

Lctrs:
Otherwise, getHistory() returns null and it's a violation of the method's contract


dunglas:
But the history isn't populated, so it's not a good fix IMHO. We should either throw when calling `getHistory()` or (better) create an implementation of `History` retrieving the needing data from WebDriver.

Lctrs:
WebDriver doesn't expose the needed data for this. So we should throw an exception. Will update the PR next week.

Lctrs:
See #355

PR #355, opened 2020-07-15
client: throw an exception when getting history

Lctrs:
WebDriver doesn't expose the needed data to make this.


Lctrs:
Failure not related

dunglas:
Thanks @Lctrs!

Issue #356, opened 2020-07-24
Install chromedriver matching installed chrome version


dbrekelmans:
Currently `chromedriver-bin/update.sh` downloads and installs the latest chromedriver.
This will cause issues in CI if you are using a built image with an older chrome version.

Instead of installing the latest version, I suggest checking the installed chrome version and downloading the matching chromedriver.

You can get the matching version like this:

```
chrome_version=$(google-chrome --version | grep -ioE "[[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+")
driver_version=$(curl -s https://chromedriver.storage.googleapis.com/LATEST_RELEASE_$chrome_version)
```

I didn't create a PR yet, because I'm unsure if getting the latest chromedriver version is a wanted feature.
Perhaps there could be 2 separate scripts or we could pass a --latest argument.

Let me know!

dunglas:
Maybe should we :

* remove the ChrimeDriver and GeckoDriver form this repository
* rewrite the shell scripts as Symfony commands (so they will work on all platforms, including Windows and BSD)
* Ask the use to run this installation command before the first run if it doesn't have ChromeDriver/GeckoDriver installed locally.

To avoir compatibility issues, we should download the exact version of ChromeDriver corresponding with the local version of Chrome (some CIs and distributions ship old versions of Chrome requiring old versions of ChromeDriver).

WDYT?

dbrekelmans:
Sounds good to me. I'll start on a draft.

Kolvin:
This would help with building docker environments too as seen [here](https://github.com/symfony/panther/issues/304#issuecomment-676055116)

dbrekelmans:
Update on the progress:

I'm working on supporting multiple OSes, browsers and drivers.

I extracted this to a separate package to better deal with the size, dependencies and testing (CI).

I will continue working on this and open a PR to use this package in panther when I have a first release ready.

dbrekelmans:
For the folks watching this thread:

If you want to help work on the solution, there are a few [easy-to-do issues](https://github.com/dbrekelmans/browser-driver-installer/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) before I can make a release. Comment on one of the issues if you want to help out 😃

PR #358, opened 2020-08-02
[DRAFT] Add install chromedriver command

dbrekelmans:
Resolves https://github.com/symfony/panther/issues/356

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,12 +18,9 @@
   ],
   "require": {
     "php": ">=7.1",
-    "ext-zip": "*",
     "php-webdriver/webdriver": "^1.8.1",
     "symfony/browser-kit": "^4.3 || ^5.0",
-    "symfony/console": "^4.3 || ^5.0",
     "symfony/dom-crawler": "^4.3 || ^5.0",
-    "symfony/filesystem": "^4.3 || ^5.0",
     "symfony/http-client": "^4.3 || ^5.0",
     "symfony/polyfill-php72": "^1.9",
     "symfony/process": "^3.4 || ^4.0 || ^5.0"
@@ -46,9 +43,12 @@
     "symfony/framework-bundle": ">=4.3 <4.3.4"
   },
   "require-dev": {
+    "ext-zip": "*",
     "fabpot/goutte": "^3.2.3",
     "guzzlehttp/guzzle": "^6.3",
+    "symfony/console": "^4.3 || ^5.0",
     "symfony/css-selector": "^3.4 || ^4.0 || ^5.0",
+    "symfony/filesystem": "^4.3 || ^5.0",
     "symfony/framework-bundle": "^3.4 || ^4.0 || ^5.0",
     "symfony/mime": "^4.3 || ^5.0",
     "symfony/phpunit-bridge": "^4.3.3 || ^5.0"
--- a/composer.json
+++ b/composer.json
@@ -18,9 +18,12 @@
   ],
   "require": {
     "php": ">=7.1",
+    "ext-zip": "*",
     "php-webdriver/webdriver": "^1.8.1",
     "symfony/browser-kit": "^4.3 || ^5.0",
+    "symfony/console": "^4.3 || ^5.0",
     "symfony/dom-crawler": "^4.3 || ^5.0",
+    "symfony/filesystem": "^4.3 || ^5.0",
     "symfony/http-client": "^4.3 || ^5.0",
     "symfony/polyfill-php72": "^1.9",
     "symfony/process": "^3.4 || ^4.0 || ^5.0"
--- a/install
+++ b/install
@@ -6,12 +6,10 @@ declare(strict_types=1);
 require __DIR__ . '/vendor/autoload.php';
 
 use Symfony\Component\Console\Application;
-use Symfony\Component\Filesystem\Filesystem;
-use Symfony\Component\HttpClient\NativeHttpClient;
 use Symfony\Component\Panther\Command\InstallChromeDriverCommand;
 
 $application = new Application();
 
-$application->add(new InstallChromeDriverCommand('chromedriver', new NativeHttpClient(), new Filesystem(), new ZipArchive()));
+$application->add(new InstallChromeDriverCommand('chromedriver'));
 
 $application->run();
--- a/install
+++ b/install
@@ -5,11 +5,38 @@ declare(strict_types=1);
 
 require __DIR__ . '/vendor/autoload.php';
 
-use Symfony\Component\Console\Application;
-use Symfony\Component\Panther\Command\InstallChromeDriverCommand;
+$dependencies = [
+    '\Symfony\Component\Console\Application',
+    '\Symfony\Component\HttpClient\NativeHttpClient',
+    '\Symfony\Component\Filesystem\Filesystem',
+];
+\array_walk(
+    $dependencies,
+    static function (string $className) : void {
+        if (!\class_exists($className)) {
+            echo \sprintf(
+                'Missing the required class "%s" to use this command. Try composer install with dev dependencies.%s',
+                $className,
+                PHP_EOL
+            );
+            exit(1);
+        }
+    }
+);
 
-$application = new Application();
+if (!\class_exists('\ZipArchive')) {
+    echo 'Missing the required php extension "ext-zip" to use this command. Make sure it is installed.' . PHP_EOL;
+    exit(1);
+}
 
-$application->add(new InstallChromeDriverCommand('chromedriver'));
+
+$application = new \Symfony\Component\Console\Application();
+
+$application->add(new \Symfony\Component\Panther\Command\InstallChromeDriverCommand(
+        'chromedriver',
+    new \Symfony\Component\HttpClient\NativeHttpClient(),
+    new \Symfony\Component\Filesystem\Filesystem(),
+    new \ZipArchive()
+));
 
 $application->run();
--- a/install
+++ b/install
@@ -1,0 +1,17 @@
+#!/usr/bin/env php
+<?php
+
+declare(strict_types=1);
+
+require __DIR__ . '/vendor/autoload.php';
+
+use Symfony\Component\Console\Application;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\HttpClient\NativeHttpClient;
+use Symfony\Component\Panther\Command\InstallChromeDriverCommand;
+
+$application = new Application();
+
+$application->add(new InstallChromeDriverCommand('chromedriver', new NativeHttpClient(), new Filesystem(), new ZipArchive()));
+
+$application->run();
--- a/src/Command/InstallChromeDriverCommand.php
+++ b/src/Command/InstallChromeDriverCommand.php
@@ -33,6 +33,9 @@ use const PHP_OS_FAMILY;
 
 final class InstallChromeDriverCommand extends Command
 {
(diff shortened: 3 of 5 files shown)


dbrekelmans:
@dunglas I made a POC, let me know if you think this is headed in the right direction. I've only tested this on linux, but I think it should work on windows and mac as well.

composer.json
```diff
@@ -18,9 +18,12 @@
   ],
   "require": {
     "php": ">=7.1",
+    "ext-zip": "*",
```
dunglas, line 21: All these new dependencies should be optional (if you use Panther as a lib, you may want to use a local ChromeDriver version, or only run this command at build time). I suggest to move these deps in the `require-dev` section and to throw an exception in the command if they are missing.

src/Command/InstallChromeDriverCommand.php
```diff
@@ -0,0 +1,400 @@
...
+     * @var Filesystem
+     */
+    private $filesystem;
+
+    /**
+     * @var ZipArchive
+     */
+    private $zip;
+
+    public function __construct(string $name, HttpClientInterface $httpClient, Filesystem $filesystem, ZipArchive $zip)
```
dunglas, line 73: These deps could be nullable and initialized directly in the constructor if not passed as parameters, so won't have to handle the DI manually in the caller script.

src/Command/InstallChromeDriverCommand.php
```diff
@@ -0,0 +1,400 @@
...
+        if ($driverVersion === self::AUTO) {
+            try {
+                $chromeVersion = $this->getInstalledChromeVersion($input->getOption(self::CHROME_BINARY));
+
+                if ($io->isVerbose()) {
+                    $io->writeln(sprintf('Chrome version "%s" found.', $chromeVersion));
+                }
+
+                $chromeDriverVersion = $this->getMatchingChromeDriverVersion($chromeVersion);
+            } catch (Throwable $exception) {
```
dunglas, line 138: This catch is a bit broad. Can we I we catch only `Exception`?

src/Command/InstallChromeDriverCommand.php
```diff
@@ -0,0 +1,400 @@
...
+                    $io,
+                    sprintf(
+                        'Could not determine chrome version. Specify the chrome binary with "--%s" or manually specify a driver version with "--%s".',
+                        self::CHROME_BINARY,
+                        self::DRIVER_VERSION
+                    ),
+                    $exception
+                );
+            }
+        } else if ($driverVersion === self::LATEST) {
```
dunglas, line 149: `elseif`

src/Command/InstallChromeDriverCommand.php
```diff
@@ -0,0 +1,400 @@
...
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output) : int
+    {
+        $io = new SymfonyStyle($input, $output);
+
+        $io->note('This command is experimental. Use at your own discretion.');
+
+        $driverVersion = $input->getOption(self::DRIVER_VERSION);
+        if ($driverVersion === self::AUTO) {
```
dunglas, line 129: You should use a `switch` here (nitpicking but it's the appropriate construct and it is a bit faster, even if it will not be noticeable here).

src/Command/InstallChromeDriverCommand.php
```diff
@@ -0,0 +1,400 @@
...
+                        self::CHROME_BINARY,
+                        self::DRIVER_VERSION
+                    ),
+                    $exception
+                );
+            }
+        } else if ($driverVersion === self::LATEST) {
+            try {
+                $chromeDriverVersion = $this->getLatestChromeDriverVersion();
+            } catch (Throwable $exception) {
```
dunglas, line 152: Same here.

src/Command/InstallChromeDriverCommand.php
```diff
@@ -0,0 +1,400 @@
...
+                return $this->fail($io, 'Unable to get the latest chrome version from API endpoint.', $exception);
+            }
+        } else {
+            try {
+                $chromeDriverVersion = $this->parseDriverVersion($driverVersion);
+            } catch (Throwable $exception) {
```
dunglas, line 158: Same here.

src/Command/InstallChromeDriverCommand.php
```diff
@@ -0,0 +1,400 @@
...
+
+        try {
+            $this->downloadChromeDriverZip(
+                $chromeDriverVersion,
+                $input->getOption(self::OS_FAMILY),
+                $io->createProgressBar()
+            );
+            $io->writeln(' Download complete.');
+
+        } catch (Throwable $exception) {
```
dunglas, line 179: Same here

src/Command/InstallChromeDriverCommand.php
```diff
@@ -0,0 +1,400 @@
...
+    }
+
+    /**
+     * @throws ProcessFailedException
+     * @throws UnexpectedValueException
+     * @throws RuntimeException
+     */
+    private function getInstalledChromeVersion(string $chromeBinary) : string
+    {
+        $process = Process::fromShellCommandline(sprintf('%s --version', $chromeBinary));
```
dunglas, line 224: It could be nice to support Chromium too.

src/Command/InstallChromeDriverCommand.php
```diff
@@ -0,0 +1,400 @@
...
+    private function getLatestChromeDriverVersion() : string
+    {
+        $response = $this->httpClient->request('GET', self::CHROMEDRIVER_API_VERSION_ENDPOINT);
+
+        return $this->parseDriverVersion($response->getContent());
+    }
+
+    private function getChromeDriverBinaryName(string $osFamily) : string
+    {
+        if ($osFamily === self::WINDOWS) {
```
dunglas, line 302: A switch should be used here too.

src/Command/InstallChromeDriverCommand.php
```diff
@@ -0,0 +1,400 @@
...
+            $progressBar->finish();
+        }
+    }
+
+    /**
+     * @throws RuntimeException
+     */
+    private function unzipChromeDriverArchive() : void
+    {
+        $success = $this->zip->open(self::DOWNLOAD_FILE_LOCATION);
```
dunglas, line 360: You can get rid of this variable, same below.

install
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env php
```
nusje2000, line 1: Since the current update command is in the chromedriver-bin directory, maybe this one should be as well?

src/Command/InstallChromeDriverCommand.php
nusje2000: This is supported by using `--chrome-binary chromium`, but maybe the correct binary could be resolved automatically?

src/Command/InstallChromeDriverCommand.php
```diff
@@ -0,0 +1,400 @@
...
+    private const AUTO = 'auto';
+    private const OS_FAMILY = 'os-family';
+    private const WINDOWS = 'Windows';
+    private const BSD = 'BSD';
+    private const DARWIN = 'Darwin';
+    private const SOLARIS = 'Solaris';
+    private const LINUX = 'Linux';
+    private const CHROMEDRIVER_API_URL = 'https://chromedriver.storage.googleapis.com';
+    private const CHROMEDRIVER_API_VERSION_ENDPOINT = self::CHROMEDRIVER_API_URL . '/LATEST_RELEASE';
+    private const BIN_DIRECTORY = __DIR__ . '/../../bin';
```
nusje2000, line 52: You could use the already existing `chromedriver-bin` directory for this.

nusje2000:
> @dunglas I made a POC, let me know if you think this is headed in the right direction. I've only tested this on linux, but I think it should work on windows and mac as well.

I'll test this branch on windows later today.

src/Command/InstallChromeDriverCommand.php
dbrekelmans: I guess we could use chromium as a fallback default if it doesn't add to much complexity. For instance, I don't know if `--version` is available in chromium or maybe it would need to be parsed differently. It's something to look into.

install
dbrekelmans: I opted to go with a general `bin/` dir so we can remove both `chromedriver-bin/` and `geckodriver-bin/` later and have the files in `bin/chromedriver` and `bin/geckodriver`

dbrekelmans:
> All these new dependencies should be optional (if you use Panther as a lib, you may want to use a local ChromeDriver version, or only run this command at build time).

@dunglas Now that I've had a day to think about it, this might be a good argument to split this into a separate chromedriver-installer package. As you've said, one should be able to use panther without this command and honestly installing the chromedriver has nothing to do with panther. WDYT?

dunglas:
I'm not sure if it is worth it tbh, it will increase the maintenance burden and make it a bit more complex for the user to install Panther. IMHO moving these new deps in the `require-dev` section and throwing with an explicit message such as "Execute `composer require --dev ...` to be able to run this command" is will be good enough for most use cases.

dunglas:
Also, could you also allow this command to download GeckoDriver? It shouldn't be too difficult regarding what you already have here (maybe adding a new parameter)?

nusje2000:
I've tested the command on windows:
 - Installing using the default strategy:
image
 - Installing using automatic version guessing:
image
 - Installing using the latest version:
image
 - Installing using a fixed version:
image

When resolving the version automatically, `google-chrome --version` is used, but this does not work on Windows. I think you have to use multiple strategies based on the OS (and if support is added, driver type). For resolving the installed version on windows you could look at [this Stack Overflow post](https://stackoverflow.com/questions/50880917/how-to-get-chrome-version-using-command-prompt-in-windows). There are lots of different methods available to resolve the version.

For additional debug information, this is the output with very verbose output:
image

dbrekelmans:
Thanks for testing @nusje2000. I will look into supporting windows as well

dunglas:
@dbrekelmans do you thin you'll have to finish this PR anytime soon? If it's not the case, can I take it over?

dbrekelmans:
@dunglas I'm actually working on a package now that could replace this. After this draft I continued working on it, but it became too large for a simple command in this repo.

https://github.com/dbrekelmans/browser-driver-installer

I mainly need to write more tests before I can create a release. When I have the first release ready I will create a PR to add it as a dev dependency here.

Surf-N-Code:
@dbrekelmans - great PR, is there any news on this? :)

dbrekelmans:
@Surf-N-Code I'm still working on the package whenever I have time (see comment above).

I've released a working version 0.1, but I think it still needs a few features before it can be usefully applied in panther.

I encourage you to take a look and contribute of you are able. I especially need some help with support for Windows.

dbrekelmans:
@dunglas Originally in this draft I directed users to the command in case they want (need) to update their driver.
`No chrome driver binary found. Use "php vendor/symfony/panther/install chromedriver" to install it.`.

My thoughts now are that we could just do it for them..
If no driver was found, we can run the [installer](https://github.com/dbrekelmans/browser-driver-installer) for them and continue with the newly installed driver.

What do you think? It would require a bit more work them simply directing them to run the installer themselves, but might be more user-friendly.

I have a release ready for google-chrome/chromium for linux/macOS. Firefox and windows not yet supported, so without those probably not ready to be included in symfony/panther, but we're getting there ;)

PR #359, opened 2020-08-27
Add waitForVisibility to wait for a present element in DOM to be visible

slimane-achbouq:
The `waitfFor()` function wait element to be present in DOM but he could be invisible so i added `waitForVisibility()` function to wait for a present element in DOM to be visible .

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -326,9 +326,7 @@ final class Client extends AbstractBrowser implements WebDriver, JavaScriptExecu
     {
         $locator = trim($locator);
 
-        $by = '' === $locator || '/' !== $locator[0]
-            ? WebDriverBy::cssSelector($locator)
-            : WebDriverBy::xpath($locator);
+        $by = $this->createWebDriverByFromLocator($locator);
 
         $this->wait($timeoutInSecond, $intervalInMillisecond)->until(
             WebDriverExpectedCondition::presenceOfElementLocated($by)
@@ -338,23 +336,24 @@ final class Client extends AbstractBrowser implements WebDriver, JavaScriptExecu
     }
 
     /**
-     * @param string $locator
-     * @param int $timeoutInSecond
-     * @param int $intervalInMillisecond
+     * @param string $locator The path to an element to be waited for. Can be a CSS selector or Xpath expression.
+     *
      * @throws NoSuchElementException
      * @throws TimeoutException
+     *
+     * @return Crawler
      */
     public function waitForVisibility(string $locator, int $timeoutInSecond = 30, int $intervalInMillisecond = 250)
     {
         $locator = trim($locator);
 
-        $by = '' === $locator || '/' !== $locator[0]
-            ? WebDriverBy::cssSelector($locator)
-            : WebDriverBy::xpath($locator);
+        $by = $this->createWebDriverByFromLocator($locator);
 
         $this->wait($timeoutInSecond, $intervalInMillisecond)->until(
             WebDriverExpectedCondition::visibilityOfElementLocated($by)
         );
+
+        return $this->crawler = $this->createCrawler();
     }
 
     public function getWebDriver(): WebDriver
@@ -526,5 +525,14 @@ final class Client extends AbstractBrowser implements WebDriver, JavaScriptExecu
         }
 
         return new WebDriverMouse($this->webDriver->getMouse(), $this);
+    }
+
+    private function createWebDriverByFromLocator(string $locator): WebDriverBy
+    {
+        $locator = trim($locator);
+
+        return '' === $locator || '/' !== $locator[0]
+            ? WebDriverBy::cssSelector($locator)
+            : WebDriverBy::xpath($locator);
     }
 }
--- a/src/Client.php
+++ b/src/Client.php
@@ -316,22 +316,42 @@ final class Client extends AbstractBrowser implements WebDriver, JavaScriptExecu
 
     /**
      * @param string $locator The path to an element to be waited for. Can be a CSS selector or Xpath expression.
-     * @param bool $checkVisibility If true means checking that an element is present on the DOM of a page and visible.
      *
      * @throws NoSuchElementException
      * @throws TimeoutException
      *
      * @return Crawler
      */
-    public function waitFor(string $locator, int $timeoutInSecond = 30, int $intervalInMillisecond = 250, bool $checkVisibility = false)
+    public function waitFor(string $locator, int $timeoutInSecond = 30, int $intervalInMillisecond = 250)
     {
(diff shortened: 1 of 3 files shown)


dunglas:
Hi, thanks for the patch, but it should already be the default behavior: https://github.com/symfony/panther/pull/236

src/Client.php
```diff
@@ -337,6 +337,26 @@ public function waitFor(string $locator, int $timeoutInSecond = 30, int $interva
         return $this->crawler = $this->createCrawler();
     }
 
+    /**
+     * @param string $locator
```
dunglas, line 341: Please remove all the `params` (we only add them when they provide additional information, and it's not the case here).

dunglas (changes requested):

src/Client.php
webignition: This param is identical in purpose to the first argument of `waitFor()`. In the case of `waitFor()`, a parameter description is added to the docblock to make it clear that this can be either a CSS selector or XPath expression.

Should the same be added here instead of removing this param?

src/Client.php
```diff
@@ -337,6 +337,26 @@ public function waitFor(string $locator, int $timeoutInSecond = 30, int $interva
...
+     * @throws NoSuchElementException
+     * @throws TimeoutException
+     */
+    public function waitForVisibility(string $locator, int $timeoutInSecond = 30, int $intervalInMillisecond = 250)
+    {
+        $locator = trim($locator);
+
+        $by = '' === $locator || '/' !== $locator[0]
+            ? WebDriverBy::cssSelector($locator)
+            : WebDriverBy::xpath($locator);
```
webignition, line 353: These lines are identical to those starting the `waitFor()` method. In this case of this method and `waitFor()` this logic always needs to be present and if it changes for one it should change for the other. 

Can this be factored out to a private method called by both `waitForVisibility()` and `waitFor()`?

webignition: Off the top of my head:

```php

private function createWebDriverByFromLocator(string $locator): WebDriverBy
{
        $locator = trim($locator);

        return '' === $locator || '/' !== $locator[0]
            ? WebDriverBy::cssSelector($locator)
            : WebDriverBy::xpath($locator);
}

File: src/Client.php [hunk 3]
```diff
@@ -337,6 +337,26 @@ public function waitFor(string $locator, int $timeoutInSecond = 30, int $interva
...
+
+        $this->wait($timeoutInSecond, $intervalInMillisecond)->until(
+            WebDriverExpectedCondition::visibilityOfElementLocated($by)
+        );
+    }
```
webignition, line 358: Does this method need to return an updated `Crawler` instance as is the case with `waitFor()`?

src/Client.php
slimane-achbouq: @dunglas @webignition i update the phpdoc of `waitForVisibility()` function to be the same of `waitfFor()` .

src/Client.php
slimane-achbouq: Thanks @webignition i added this function and use it in both `waitForVisibility()` and `waitFor()` .

src/Client.php
slimane-achbouq: I think is not necessary but i updated the function to return the updated `Crawler` .

webignition: Just to be clear, I wasn't suggesting that was necessary.

 I noticed there was a discrepancy between two methods that were almost identical.

webignition: But if it is acceptable for this method to return the updated crawler, there is now just a single line difference between `waitForVisibility()` and `waitFor()` which could probably be further factored out.

src/Client.php
```diff
@@ -316,23 +316,22 @@ public function getCookieJar()
...
      * @param string $locator The path to an element to be waited for. Can be a CSS selector or Xpath expression.
+     * @param bool $checkVisibility If true means checking that an element is present on the DOM of a page and visible.
      *
      * @throws NoSuchElementException
      * @throws TimeoutException
      *
      * @return Crawler
      */
-    public function waitFor(string $locator, int $timeoutInSecond = 30, int $intervalInMillisecond = 250)
+    public function waitFor(string $locator, int $timeoutInSecond = 30, int $intervalInMillisecond = 250, bool $checkVisibility = false)
```
webignition, line 326: The previous implementation, of having two methods (`waitFor()` and `waitForVisibility()`) is a better choice than the addition of a boolean parameter.

With the introduction of a boolean parameter, this method now does not one thing but two things with the boolean parameter letting you choose which of the two things to execute. A method should do just one thing.

From the calling perspective, the downside to a boolean parameter is that the purpose is not clear when reading code. This can lead to confusing interpretations, make code harder to reason about and makes it easier to introduce bugs.

For the sake of an example, let's simplify things and assume that `waitFor()` takes only two parameters:

```php

public function waitFor(string $locator, bool $checkVisibility): Crawler
{
    // ...
}
```

From reading code that uses this method, it is not possible to determine, without examining the method definition, the difference between these three calls:

```php

$client->waitFor('.selector');
$client->waitFor('.selector', false);
$client->waitFor('.selector', true);
```

Two separate public methods with clear names remove confusion. The method names alone make it clear that there are two different methods achieving two different things:
```php

$client->waitFor('.selector');
$client->waitForVisibility('.selector');
```

src/Client.php
```diff
@@ -326,9 +326,7 @@ public function waitFor(string $locator, int $timeoutInSecond = 30, int $interva
     {
         $locator = trim($locator);
```
dunglas, line 327:
```suggestion
```

`trim()` is already called in `createWebDriverByFromLocator()`

src/Client.php
```diff
@@ -337,6 +335,27 @@ public function waitFor(string $locator, int $timeoutInSecond = 30, int $interva
...
+     * @param string $locator The path to an element to be waited for. Can be a CSS selector or Xpath expression.
+     *
+     * @throws NoSuchElementException
+     * @throws TimeoutException
+     *
+     * @return Crawler
+     */
+    public function waitForVisibility(string $locator, int $timeoutInSecond = 30, int $intervalInMillisecond = 250)
+    {
+        $locator = trim($locator);
```
dunglas, line 348:
```suggestion
```

same here

dunglas (changes requested):

src/Client.php
slimane-achbouq: Removed, thanks

src/Client.php
slimane-achbouq: done !

dunglas (approved):

dunglas:
Thanks @slimane-achbouq

PR #363, opened 2020-09-10
Disable process output to stop timeout error

JohnstonCode:
"Operation timed out after 30000 milliseconds with 0 bytes received"

I was getting this error when running all tests that use Panther.

After some trial and error i found that disabling the output for the process created by WebServerManager seemed to resolve the issue.

If i run each test individually or a small selection i don't get this error.

**Update**
After some testing it appear that the issue is caused but the stderr pipe not being emptied. This is then causing the php server process to lock up. You can either empty the stderr pipe or set it to `fopen('/dev/null', 'c')` to remedy the situation.

This change will not affect anything as only the status of the process is queried.

### DIFF
--- a/src/ProcessManager/WebServerManager.php
+++ b/src/ProcessManager/WebServerManager.php
@@ -71,6 +71,8 @@ final class WebServerManager
             null,
             null
         );
+        // Stop Operation timed out after 30001 milliseconds with 0 bytes received errors
+        $this->process->disableOutput();
 
         // Symfony Process 3.4 BC: In newer versions env variables always inherit,
         // but in 4.4 inheritEnvironmentVariables is deprecated, but setOptions was removed


webignition:
How does this resolve the issue?

Previously an error was presented when an operation timed out. Does the operation now not time out? Or does it still time out and no timeout error is shown?

JohnstonCode:
I'm getting this error when i run all testing using Panther "Operation timed out after 30001 milliseconds with 0 bytes received". 

It fails on the same test every time. If i remvoe the test it just fails on the next one. If i run the server manually using `php -dvariables_order=EGPCS -S 127.0.0.1:9080 -t public/` All the tests run and pass. If i disable the process output in WebServerManager all the tests run and there is no error. If i make the process public and add this to my code all the test run and pass
```
protected function tearDown() : void
{
	var_dump(self::$webServerManager->process->getIncrementalErrorOutput());
}
```
Is there a better way i can find out what is causing the process created to hang?

JohnstonCode:
Could be this because the pipe is getting blocked because stdout and stderr are not being emptied?

JohnstonCode:
if i overwrite https://github.com/symfony/process/blob/master/Process.php#L292 with
```
$descriptors = [
	['pipe', 'r'],
    ['pipe', 'w'], // stdout
    ["file", "/tmp/error-output.txt", "a"], // stderr
];
```
All the tests run

JohnstonCode:
the reason it works with `disableOutput` is the descriptors are
```
$nullstream = fopen('/dev/null', 'c');
return [
	['pipe', 'r'],
	$nullstream,
	$nullstream,
];
```

webignition:
I don't know anywhere near enough of how the current code works to even start suggesting why things may fail under certain conditions and how things may work and/or appear to work with the described changes.

My concern is that the PR description states that this solution was arrived at through trial and error. If I were reviewing a PR for a colleague and was presented with similar reasons for making a change, I'd not be confident that the root cause was fully understood nor that the side effects (if any) of the change were appreciated.

PR #365, opened 2020-09-17
Disable web server process output to stop operation timeout

### LINKED ISSUE #364: Operation timed out after 30000 milliseconds with 0 bytes received
I am getting this error when running multiple Panther tests. `Operation timed out after 30000 milliseconds with 0 bytes received`

This issue does not occur when running tests individually or not using the `Symfony\Component\Panther\ServerExtension` extension.

The issue also doesn't occur when running the server manually with `php -dvariables_order=EGPCS -S 127.0.0.1:9080 -t public/`.

After some digging i believe that the issue is caused by the stderr pipe not being emptied and causing the server process to lock up. This issue will not affect windows users as [they use temporary files instead of pipes](https://github.com/symfony/process/blob/master/Pipes/WindowsPipes.php#L43).

If you either set the [Process](https://github.com/symfony/panther/blob/master/src/ProcessManager/WebServerManager.php#L56) to `disableOutput` which sets the [descriptor](https://github.com/symfony/process/blob/6beca1cfc85d432ba66be4233ffb9ce4386ce85c/Pipes/UnixPipes.php#L49) of stderr to be `fopen('/dev/null', 'c')` then the issue is no longer present.

Also if you set the [Process](https://github.com/symfony/process/blob/master/Process.php#L292) descriptor to
```
$descriptors = [
	['pipe', 'r'],
	['pipe', 'w'], // stdout
	['file', '/tmp/error-output.txt', 'a'], // stderr
];
```
the issue is no longer present.

You can also run `getIncrementalErrorOutput` or `getErrorOutput` and this will stop the error from happening also.

I think setting `$this->process->disableOutput();` just after https://github.com/symfony/panther/blob/master/src/ProcessManager/WebServerManager.php#L73 should resolve this issue and not affect anything else. This is because only the status of the web server process is checked and the output is not used.

JohnstonCode:
This PR disables the output from the web server process to stop operation time-outs happening. The web server process is locking as the stderr pipe is not getting emptied.

fixes #364

### DIFF
--- a/src/ProcessManager/WebServerManager.php
+++ b/src/ProcessManager/WebServerManager.php
@@ -71,6 +71,7 @@ final class WebServerManager
             null,
             null
         );
+        $this->process->disableOutput();
 
         // Symfony Process 3.4 BC: In newer versions env variables always inherit,
         // but in 4.4 inheritEnvironmentVariables is deprecated, but setOptions was removed


dunglas (approved):

chalasr (approved):

dunglas:
Thanks @JohnstonCode

PR #366, opened 2020-09-18
WebDriverCookie "secure" flag read from WebDriverCookie

codegain:
In the function `symfonyToWebDriver` in `CookieJar.php` the value `secure` of a cookie is read from `$webDriverCookie` and then also set on `$webDriverCookie`. I think this should be read from the symfony `$cookie` variable.


dunglas:
Thanks @codegain

PR #369, opened 2020-09-24
Bump php-webdriver to ^1.8.2 and symfony/* to ^4.4

chalasr:


dunglas:
There is still one test failing in lowest mode.

chalasr:
@dunglas Now green.

dunglas:
Awesome! Thank you @chalasr

PR #370, opened 2020-09-25
browser manager options

### LINKED ISSUE #368: additional web driver capabilities
currently it seems impossible to pass additional capabilities to the web driver.

in development situations it's not uncommon to work with self signed ssl certificates, but without being able to set the ``acceptInsecureCerts`` capability to ``true`` you will always run into the browser's 'not secure' warning when running a scenario.

one way of getting around this could be passing a 'capabilities' array in the options argument of the browser manager's constructor and apply those in the start method.

any thoughts?

wickedOne:
provides a way to pass browser manager options

fixes #368

### DIFF
--- a/src/PantherTestCaseTrait.php
+++ b/src/PantherTestCaseTrait.php
@@ -160,7 +160,7 @@ trait PantherTestCaseTrait
      *
      * @param array $options see {@see $defaultOptions}
      */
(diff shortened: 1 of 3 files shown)


dunglas:
Thanks @wickedOne

PR #372, opened 2020-10-02
ci: switch to GitHub Actions

dunglas:
* Switch from Travis and AppVeyor to GitHub Actions
* Run tests on Mac OS with Firefox and Chrome


.github/workflows/ci.yml
```diff
@@ -0,0 +1,177 @@
...
+                  restore-keys: ${{ runner.os }}-composer-
+
+            - name: Install dependencies
+              run: composer install --prefer-dist
+
+            - name: Install PHPUnit dependencies
+              run: vendor/bin/simple-phpunit --version
+
+            - name: PHPStan
+              run: phpstan analyse --no-progress --error-format=checkstyle | cs2pr
```
stof, line 55: phpstan has native support for Github Actions. So using `phpstan analyse --no-progress` would be better:

- you would still have annotations
- you would have the full phpstan report in a human readable way in the logs of the action, which is relevant when there is lots of reported issues (github actions are limited in the number of annotations they can report)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,177 @@
...
+
+    phpunit:
+        runs-on: ubuntu-latest
+        strategy:
+            matrix:
+                php-versions: ['7.1', '7.2', '7.3', '7.4']
```
stof, line 61: I suggest adding `fail-fast: false` in the strategy too, so that tests run on the 4 PHP versions all the time, rather than being cancelled when one of the fail

.github/workflows/ci.yml
```diff
@@ -0,0 +1,177 @@
...
+
+            - name: Get composer cache directory
+              id: composercache
+              run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+
+            - name: Cache dependencies
+              uses: actions/cache@v2
+              with:
+                  path: ${{ steps.composercache.outputs.dir }}
+                  key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.json') }}
```
bendavies, line 80: this should be `hashFiles('composer.json')`, or you are generating a hash on all the composer.json files in the `vendor` directory when the cache is saved, creating a different key than the restoration key where only the root `composer.json` will be considered.

PR #376, opened 2020-10-06
Fix baseUri not taken into account when using Symfony HttpBrowser

### LINKED ISSUE #313: Symfony HttpBrowser ignores PANTHER_EXTERNAL_BASE_URI
Hello,

On our current project, we are replacing (deprecated) Goutte clients by HttpBrowser ones.
Some tests are failing ; it appears that `PANTHER_EXTERNAL_BASE_URI` env variable is not respected. Requests without URI will be send to `http://localhost`.

Upon investigation, it seems that HttpBrowser ignores `baseUri` (from https://github.com/symfony/symfony/pull/35177).
It prevents (strict) interoperability between clients. In our use case, we wanted to use Chrome for tests needing javascript, and HttpBrowser for the rest, which is a quite common use case I believe.

I don't know what to do from there: decorate HttpBrowser?
At least this limitation should be documented.

lolautruche:
By default, HttpBrowser expects absolute URLs, but this makes tests interoperability almost impossible.
However, it is possible to tell HttpBrowser to use a given host, by setting the `HTTP_HOST` server variable, in conjunction with `HTTPS` if applicable.

This change ensures that `self::$baseUri` is correctly taken into
account in this regard.

Should fix #313

### DIFF
--- a/src/PantherTestCaseTrait.php
+++ b/src/PantherTestCaseTrait.php
@@ -245,6 +245,12 @@ trait PantherTestCaseTrait
             static::bootKernel($kernelOptions); // @phpstan-ignore-line
         }
 
+        $urlComponents = parse_url(self::$baseUri);
+        self::$httpBrowserClient->setServerParameter('HTTP_HOST', sprintf('%s:%s', $urlComponents['host'], $urlComponents['port']));
+        if ('https' === $urlComponents['scheme']) {
+            self::$httpBrowserClient->setServerParameter('HTTPS', 'true');
+        }
+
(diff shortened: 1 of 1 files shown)


lolautruche:
Looks like the latest check was cancelled

dunglas (approved):

lolautruche:
There is still an issue with PHP 7.2 on ubuntu-latest. Issue related to ChromeDriver, not to this PR

dunglas:
Thanks @lolautruche!

PR #377, opened 2020-10-13
Fix symfony/phpunit-bridge version

laurent35240:
When trying to install panther locally I ended up with this composer error
```
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - The requested package symfony/phpunit-bridge dev-master exists as symfony/phpunit-bridge[2.7.x-dev, ..., v5.2.0-BETA1] but these are rejected by your constraint.
```

I found out that there is no more branch `master` on `symfony/phpunit-bridge` project but branch `5.x` is the default one now


dunglas (approved):

dunglas:
Thanks @laurent35240!

PR #385, opened 2020-11-16
feat: add assertion about visibility/enabled status and waitFor methods for invisibility/staleness and text to be present or not

Gregcop1:
**assertions**
- feat: add assertion `assertSelectorIsDisabled`
- feat: add assertion `assertSelectorIsNotVisible`

**client waitFor methods**
- feat: add client method `waitForStaleness`  (element is removed from the DOM)
- feat: add client method `testWaitForInvisibility` (element disappears)
- feat: add client method `waitForElementToContain` (a text is inserted in element's content)
- feat: add client method `waitForElementToNotContain` (a text is removed from element's content)


README.md
```diff
@@ -125,8 +125,23 @@ class E2eTest extends PantherTestCase
         $client->request('GET', '/mypage');
 
         // Use any PHPUnit assertion, including the ones provided by Symfony
-        $this->assertPageTitleContains('My Title');
-        $this->assertSelectorTextContains('#main', 'My body');
+        self::assertPageTitleContains('My Title');
```
dunglas, line 128: We use `$this->` in Symfony: https://symfony.com/doc/current/testing/functional_tests_assertions.html

src/Client.php
```diff
@@ -524,7 +586,7 @@ public function getMouse(): WebDriverMouse
         return new WebDriverMouse($this->webDriver->getMouse(), $this);
     }
 
-    private function createWebDriverByFromLocator(string $locator): WebDriverBy
+    public static function createWebDriverByFromLocator(string $locator): WebDriverBy
```
dunglas, line 589: I would move this method to an internal trait or class instead. But marking it `@internal` is maybe enough.

src/WebDriver/WebDriverExpectedCondition.php
```diff
@@ -0,0 +1,45 @@
...
+     * @return static condition returns whether the partial text is present in the element
+     */
+    public static function elementTextNotContains(WebDriverBy $by, string $text)
+    {
+        return new static(
+            function (WebDriver $driver) use ($by, $text) {
+                try {
+                    $element_text = $driver->findElement($by)->getText();
+
+                    return false === mb_strpos($element_text, $text);
```
dunglas, line 38: Isn't `strpos` enough? Otherwise, we'll have to add `symfony/polyfill-mbstring` as dependency in case the extension isn't available. It's not a problem because it's already a dep of PHP WebDriver, but we must also list it to avoid transient dependencies.

src/WebDriver/WebDriverExpectedCondition.php
```diff
@@ -0,0 +1,45 @@
...
+declare(strict_types=1);
+
+namespace Symfony\Component\Panther\WebDriver;
+
+use Facebook\WebDriver\Exception\StaleElementReferenceException;
+use Facebook\WebDriver\WebDriver;
+use Facebook\WebDriver\WebDriverBy;
+use Facebook\WebDriver\WebDriverExpectedCondition as BaseWebDriverExpectedCondition;
+
+final class WebDriverExpectedCondition extends BaseWebDriverExpectedCondition
```
dunglas, line 21: Extending `BaseWebDriverExpectedCondition` isn't necessary, isn't it? Using composition is preferable.

tests/AssertionsTest.php
```diff
@@ -32,20 +32,26 @@ protected function setUp(): void
     public function testDomCrawlerAssertions(): void
     {
         self::createPantherClient()->request('GET', '/basic.html');
-        $this->assertSelectorExists('.p-1');
```
dunglas, line 35: These changes should be reverted for consistency with the rest of the Symfony codebase.

dunglas (changes requested):
Nice set of new assertions! Thank you for contributing this.

README.md
Gregcop1: Ok to change but can you explain me why this choice since those function are `static` ?

dunglas: It's historical IIRC.

dunglas (approved):

dunglas:
Thanks @Gregcop1

PR #388, opened 2020-12-02
Prevent PHP 8 from outputting fatal error when calling fsockopen()

jtraulle:
Since PHP 8, @ Error Suppression operator does not silent fatal errors anymore.
See https://php.watch/versions/8.0/fatal-error-suppression for more.

This change prevent PHP 8 from outputting
Warning Error: fsockopen(): Unable to connect to 127.0.0.1:9515 (Connection refused)

### DIFF
--- a/src/ProcessManager/WebServerReadinessProbeTrait.php
+++ b/src/ProcessManager/WebServerReadinessProbeTrait.php
@@ -29,7 +29,10 @@ trait WebServerReadinessProbeTrait
      */
     private function checkPortAvailable(string $hostname, int $port, bool $throw = true): void
     {
-        $resource = @fsockopen($hostname, $port);
+        $current_state = error_reporting();
+        error_reporting(0);
+        $resource = fsockopen($hostname, $port);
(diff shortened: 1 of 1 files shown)


src/ProcessManager/WebServerReadinessProbeTrait.php
```diff
@@ -29,7 +29,10 @@ trait WebServerReadinessProbeTrait
      */
     private function checkPortAvailable(string $hostname, int $port, bool $throw = true): void
     {
-        $resource = @fsockopen($hostname, $port);
+        $current_state = error_reporting();
+        error_reporting(0);
+        $resource = fsockopen($hostname, $port);
+        error_reporting($current_state);
```
dunglas, line 35:
```suggestion
        $currentState = error_reporting();
        error_reporting(0);
        $resource = fsockopen($hostname, $port);
        error_reporting($currentState);
```

dunglas (changes requested):

dunglas:
Thanks @jtraulle

PR #389, opened 2020-12-02
Add composer script to easily update drivers

dbrekelmans:
Adds a `composer update-drivers` script to update the chromedriver and geckodriver binaries to the latest version. Also suggests [dbrekelmans/bdi](https://github.com/dbrekelmans/bdi) to users if they don't want to use the latest drivers.

You could add this as a CI job to run periodically and automatically commit the new binaries, which would remove some maintenance for you.

This is a first step, but the end goal is still https://github.com/symfony/panther/pull/358#issuecomment-716663609


kbond:
This library is indeed useful, Panther isn't currently testing on PHP 8 but what's the status of https://github.com/dbrekelmans/browser-driver-installer/issues/39?

dunglas:
Nice! Shouldn't we remove the binary included in this repo and recommend using the command if there is no existing global install of ChromeDriver/Geckodriver? WDYT @nicolas-grekas @fabpot?

dbrekelmans:
> This library is indeed useful, Panther's CI isn't currently testing on PHP 8 but what's the status of https://github.com/dbrekelmans/browser-driver-installer/issues/39?

I have not yet tested it on PHP 8, but I suppose it will be compatible with no or minor changes. Needs a PR to add CI on PHP 8 and change the composer.json and fix anything that breaks.

It also needs some changes for PHP 7.1 compatibility unless panther drops support for 7.1.

I might look into it this weekend, but open for PRs of course!

dbrekelmans:
> Nice! Shouldn't we remove the binary included in this repo and recommend using the command if there is no existing global install of ChromeDriver/Geckodriver? WDYT @nicolas-grekas @fabpot?

I thought maybe to leave them in now as a fallback until the tool is proven to be stable enough to be fully relied on.

chromedriver-bin/update.sh
```diff
@@ -1,22 +0,0 @@
-#!/usr/bin/env bash
```
nusje2000, line 1: Since we are only suggesting the installation of the bdi package, maybe we should keep this script for everyone who does not need the added features provided by the bdi package.

composer.json
```diff
@@ -48,6 +48,20 @@
...
+    "dbrekelmans/bdi": "^0.2"
+  },
+  "scripts": {
+    "update-drivers": [
+      "vendor/bin/bdi driver:chromedriver chromedriver-bin --os=windows",
+      "vendor/bin/bdi driver:chromedriver chromedriver-bin --os=macos && mv chromedriver-bin/chromedriver chromedriver-bin/chromedriver_mac64",
+      "vendor/bin/bdi driver:chromedriver chromedriver-bin --os=linux && mv chromedriver-bin/chromedriver chromedriver-bin/chromedriver_linux64",
+      "vendor/bin/bdi driver:geckodriver geckodriver-bin --os=windows",
+      "vendor/bin/bdi driver:geckodriver geckodriver-bin --os=macos && mv geckodriver-bin/geckodriver geckodriver-bin/geckodriver_mac64",
+      "vendor/bin/bdi driver:geckodriver geckodriver-bin --os=linux && mv geckodriver-bin/geckodriver geckodriver-bin/geckodriver_linux64"
```
nusje2000, line 61: Haven't tested this on windows yet, but if I'm correct, windows does not have a mv command, so this won't work on windows. Windows does have a `move <origin> <destination>` command.

nusje2000 (approved):

composer.json
dbrekelmans: This is a command for the panther maintainers, not end user. Therefor I doubt windows support is required

chromedriver-bin/update.sh
dbrekelmans: The package replaces exactly this update script. The "added features" are multi OS and browser support and some command line options

dbrekelmans:
@dunglas I'm having some issues supporting PHP7.1, especially with building the PHAR. It's more hassle than I think it's worth. What are the chances of panther dropping support for 7.1? It's been EOL for over a year now anyway: https://www.php.net/supported-versions.php and only 7.45% of people use it: https://blog.packagist.com/php-versions-stats-2020-2-edition/

dunglas:
@dbrekelmans Symfony 5 requires PHP >=7.2.5, so I'm fine with dropping PHP 7.1.

dbrekelmans:
I don't think the CI failures are related to this PR.

PR #391, opened 2020-12-07
Screenshot improvements

kbond:
This feature allow setting a base screenshot directory via the `PANTHER_SCREENSHOT_DIR` env variable. When this is set, you can exclude the absolute path when calling `$client->takeScreenshot()`.

Example, assuming the `PANTHER_SCREENSHOT_DIR` has been set to `./screenshots`:

```php
$client->takeScreenshot('about.png'); // saves to ./screenshots/about.png
$client->takeScreenshot('/contact.png'); // saves to ./screenshots/contact.png
```

Issue #73, opened 2018-07-30
take screenshot of failed assertions


ChangePlaces:
When a failed assertion occurs, how about saving the screen with the current timestamp?

dunglas:
Would be great! Do you want to work on this?

ChangePlaces:
panther has no use for me now!

nicolas-grekas:
Closing then, no need to keep a wish list.

javiereguiluz:
@ChangePlaces I don't know you, so maybe there's a language barrier issue (or maybe you are native English person, I don't know) but I'd like you to please publish more gentle comments.

This is an open source project, so the work is volunteered by people. Comments like this one or the others you published (https://github.com/symfony/panther/issues/60#issuecomment-42098, https://github.com/symfony/panther/issues/62#issuecomment-420989339) don't help at all to the project. Thanks a lot!

PR #392, opened 2020-12-07
[Feature] Automatically take screenshot on failure/error

kbond:
When using the PHPUnit extension and have `PANTHER_SCREENSHOT_DIR` configured, failures and errors for tests that are using the `PantherClient` will have screenshots saved in `PANTHER_SCREENSHOT_DIR` in the format of `{timestamp}_{error|failure}_Name-Space-MyTest__test_my_test.png`.

This is a possible solution to #73 and relies on #391. I wasn't sure the best way to allow this and am open to alternative suggestions. I'm also not sure how/if we can test this. I didn't add this feature to the `ServerListener` but can if desired.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,10 @@ If you intend to use Panther to test your application, we strongly recommend reg
 While not strictly mandatory, this extension dramatically improves the testing experience by boosting the performance and
 allowing to use the [interactive debugging mode](#interactive-mode).
 
+When using the extension in conjunction with `PANTHER_SCREENSHOT_DIR` environment variable, test using the Panther
+client that fail or error (after the client is created) will get automatically get a screenshot taken to help
+debug.
+
 To register the Panther extension, add the following lines to `phpunit.xml.dist`:
 
 ```xml
--- a/README.md
+++ b/README.md
@@ -76,9 +76,9 @@ If you intend to use Panther to test your application, we strongly recommend reg
 While not strictly mandatory, this extension dramatically improves the testing experience by boosting the performance and
 allowing to use the [interactive debugging mode](#interactive-mode).
 
-When using the extension in conjunction with `PANTHER_SCREENSHOT_DIR` environment variable, test using the Panther
-client that fail or error (after the client is created) will get automatically get a screenshot taken to help
-debug.
+When using the extension in conjunction with the `PANTHER_SCREENSHOT_DIR` environment variable, tests using the Panther
+client that fail or error (after the client is created) will automatically get a screenshot taken to help
+debugging.
 
 To register the Panther extension, add the following lines to `phpunit.xml.dist`:
 
--- a/README.md
+++ b/README.md
@@ -339,6 +339,7 @@ The following environment variables can be set to change some Panther's behaviou
 * `PANTHER_WEB_SERVER_ROUTER`:  to use a web server router script which is run at the start of each HTTP request
 * `PANTHER_EXTERNAL_BASE_URI`: to use an external web server (the PHP built-in web server will not be started)
 * `PANTHER_APP_ENV`: to override the `APP_ENV` variable passed to the web server running the PHP app
+* `PANTHER_SCREENSHOT_DIR`: to set a base directory for your screenshots (ie `./screenshots`)
 
 ### Changing the Hostname and Port of the Built-in Web Server
 
--- a/README.md
+++ b/README.md
@@ -343,7 +343,7 @@ The following environment variables can be set to change some Panther's behaviou
 * `PANTHER_WEB_SERVER_ROUTER`:  to use a web server router script which is run at the start of each HTTP request
 * `PANTHER_EXTERNAL_BASE_URI`: to use an external web server (the PHP built-in web server will not be started)
 * `PANTHER_APP_ENV`: to override the `APP_ENV` variable passed to the web server running the PHP app
-* `PANTHER_SCREENSHOT_DIR`: to set a base directory for your screenshots (ie `./screenshots`)
+* `PANTHER_SCREENSHOT_DIR`: to set a base directory for your screenshots (e.g. `./screenshots`)
 
 ### Changing the Hostname and Port of the Built-in Web Server
 
--- a/README.md
+++ b/README.md
@@ -76,8 +76,8 @@ If you intend to use Panther to test your application, we strongly recommend reg
 While not strictly mandatory, this extension dramatically improves the testing experience by boosting the performance and
 allowing to use the [interactive debugging mode](#interactive-mode).
 
-When using the extension in conjunction with the `PANTHER_SCREENSHOT_DIR` environment variable, tests using the Panther
-client that fail or error (after the client is created) will automatically get a screenshot taken to help
+When using the extension in conjunction with the `PANTHER_ERROR_SCREENSHOT_DIR` environment variable, tests using the
+Panther client that fail or error (after the client is created) will automatically get a screenshot taken to help
 debugging.
 
 To register the Panther extension, add the following lines to `phpunit.xml.dist`:
@@ -343,7 +343,7 @@ The following environment variables can be set to change some Panther's behaviou
 * `PANTHER_WEB_SERVER_ROUTER`:  to use a web server router script which is run at the start of each HTTP request
 * `PANTHER_EXTERNAL_BASE_URI`: to use an external web server (the PHP built-in web server will not be started)
 * `PANTHER_APP_ENV`: to override the `APP_ENV` variable passed to the web server running the PHP app
-* `PANTHER_SCREENSHOT_DIR`: to set a base directory for your screenshots (e.g. `./screenshots`)
+* `PANTHER_ERROR_SCREENSHOT_DIR`: to set a base directory for your failure/error screenshots (e.g. `./screenshots`)
 
 ### Changing the Hostname and Port of the Built-in Web Server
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -52,6 +52,10 @@ If you intend to use Panther to test your application, we strongly recommended t
 While not strictly mandatory, this extension dramatically improves the testing experience by boosting the performance and
 allowing to use the [interactive debugging mode](#interactive-mode).
 
+When using the extension in conjunction with `PANTHER_SCREENSHOT_DIR` environment variable, test using the Panther
```
dunglas, line 55:
```suggestion
When using the extension in conjunction with the `PANTHER_SCREENSHOT_DIR` environment variable, tests using the Panther
```

README.md
```diff
@@ -52,6 +52,10 @@ If you intend to use Panther to test your application, we strongly recommended t
...
+client that fail or error (after the client is created) will get automatically get a screenshot taken to help
```
dunglas, line 56:
```suggestion
client that fail or error (after the client is created) will automatically get a screenshot taken to help
```

README.md
```diff
@@ -52,6 +52,10 @@ If you intend to use Panther to test your application, we strongly recommended t
...
+debug.
```
dunglas, line 57:
```suggestion
debugging.
```

src/Client.php
```diff
@@ -436,6 +436,10 @@ public function takeScreenshot($saveAs = null)
     {
         $this->start();
 
+        if ($saveAs && ($dir = $_SERVER['PANTHER_SCREENSHOT_DIR'] ?? false)) {
```
dunglas, line 439: Shouldn't you test if it's not already an absolute or relative path that is passed?

src/PantherTestCaseTrait.php
```diff
@@ -92,6 +92,11 @@ public static function tearDownAfterClass(): void
         }
     }
 
+    public static function getPrimaryPantherClient(): ?PantherClient
```
dunglas, line 95: Why this change? I'm not a fond of adding this method.

src/ServerTrait.php
```diff
@@ -44,4 +44,13 @@ private function pause($message): void
...
         }
     }
+
+    private function takeScreenshot(string $type, string $test): void
+    {
+        if (!($_SERVER['PANTHER_SCREENSHOT_DIR'] ?? false) || !$client = PantherTestCase::getPrimaryPantherClient()) {
+            return;
+        }
+
+        $client->takeScreenshot(sprintf('%s_%s.png', $type, strtr($test, ['\\' => '-', ':' => '_'])));
```
dunglas, line 54: I suggest prepending the path by the current date and time.

dunglas (changes requested):
Nice feature! I can't wait to merge this.

src/Client.php
kbond: The directory might not exist so I can't check that - are you thinking just to look if the string starts with a `.`?

src/PantherTestCaseTrait.php
kbond: I don't like it either but was the only way I could think of to get the current client in the PHPUnit extension. In [zenstruck/browser](https://github.com/zenstruck/browser), I sort of do the opposite: when the browser/clients are created, I [register with the extension](https://github.com/zenstruck/browser/blob/25b111e121abb38a91eb587434828bb987dc512e/src/Browser/Test/HasBrowser.php#L57). Would you like something like this or do you have another suggestion?

src/ServerTrait.php
kbond: I thought of this too but decided against this as it could make your screenshot directory huge when running the test suite many times with many failures. I figured letting them override the old files would be best.

That being said, I'll prepend the date/time if you'd like. Just wanted to give my reasoning for not doing it.

src/Client.php
dunglas: You can use this helper: https://symfony.com/doc/current/components/filesystem.html#isabsolutepath

src/ServerTrait.php
dunglas: I fear it can make things hard to debug in case of override. I prefer to save everything and clean after than loosing some maybe important debugging data.

src/PantherTestCaseTrait.php
dunglas: I definitely prefer how you do in Browser! Can we switch to something like this?

kbond: Sounds good. Another thing I do in browser is take screenshots for all browser instances used in the tests. I should do the same thing here, yes?

src/Client.php
kbond: What if someone has the `$_SERVER['PANTHER_SCREENSHOT_DIR']` set and they pass an "absolute" path (ie `takeScreenshot('/screenshot.jpg')`)?

src/ServerTrait.php
kbond: See 6a23c013bf08297d5b235afe5dbe35f6026818af, let me know if you'd like a different format.

src/PantherTestCaseTrait.php
dunglas: Good idea!

kbond: See 113fafe2fa5fb31c4f2bfe39c0e052ce695dedca, wdyt?

src/Client.php
dunglas: Then the screenshot should be stored at `/` as it's an absolute path, isn't it?

src/PantherTestCaseTrait.php
dunglas: LGTM!

README.md
```diff
@@ -294,6 +298,7 @@ The following environment variables can be set to change some Panther's behaviou
 * `PANTHER_WEB_SERVER_ROUTER`:  to use a web server router script which is run at the start of each HTTP request
 * `PANTHER_EXTERNAL_BASE_URI`: to use an external web server (the PHP built-in web server will not be started)
 * `PANTHER_APP_ENV`: to override the `APP_ENV` variable passed to the web server running the PHP app
+* `PANTHER_SCREENSHOT_DIR`: to set a base directory for your screenshots (ie `./screenshots`)
```
dunglas, line 301:
```suggestion
* `PANTHER_SCREENSHOT_DIR`: to set a base directory for your screenshots (e.g. `./screenshots`)
```

src/Client.php
kbond: Yep ok, I currently treat [this as relative](https://github.com/symfony/panther/pull/392/files#diff-ce3bd06cad1893ba2abb840a3061bb39e857377c5feeddfdbdee1bd03b51be81R68). I'll make the change.

src/Client.php
```diff
@@ -499,7 +500,15 @@ public function takeScreenshot($saveAs = null)
     {
         $this->start();
 
-        return $this->webDriver->takeScreenshot($saveAs);
+        if (null === $saveAs || (new Filesystem())->isAbsolutePath($saveAs)) {
+            return $this->webDriver->takeScreenshot($saveAs);
+        }
+
+        if (!($dir = $_SERVER['PANTHER_SCREENSHOT_DIR'] ?? false)) {
+            throw new \RuntimeException('$saveAs must be an absolute path if not using the PANTHER_SCREENSHOT_DIR env variable.');
```
dunglas, line 508: What's the point of throwing here? Shouldn't we keep the existing behavior? (save to the current wording directory)?

Actually, I wonder if we shouldn't keep the code in `Client` as-is and support this env var only for tests. We could rename the env var to `PANTHER_ERROR_SCREENSHOT_DIR` to be more explicit. WDYT?

dunglas (changes requested):

src/Client.php
kbond: What about `Client::setScreenshotDir()` (see https://github.com/symfony/panther/pull/391#issuecomment-740593928). I think it could be useful for users to have a known directory for all the screenshots they manually take in their tests.

nicolas-grekas:
Can you please rebase on latest `main`? Becareful: `main` has been force-pushed.

nicolas-grekas:
(and remove the POC prefix is it's ready for merge ;) )

kbond:
I've rebased and removed the POC prefix but I'm thinking https://github.com/symfony/panther/pull/392#pullrequestreview-564907313 should be resolved before merging.

src/Client.php
dunglas: But when using the standalone library (to create a web crawler for instance), saving relatively to the current working directory (as currently) looks like the obvious and expected behavior to me. Having to pass an absolute path or to set an env var is cumbersome.

nicolas-grekas: I also think this check makes no sense.
Also, I think one should be able to configure the client to *not* use the env var.
Can't we use an option? I see that the constructor takes an `$options` array.

src/Client.php
```diff
@@ -509,7 +510,15 @@ public function takeScreenshot($saveAs = null)
     {
         $this->start();
 
-        return $this->webDriver->takeScreenshot($saveAs);
+        if (null === $saveAs || (new Filesystem())->isAbsolutePath($saveAs)) {
```
nicolas-grekas, line 513: I mean this one: why would we force one to pass an absolute path? I don't understand.

src/Client.php
kbond: > Actually, I wonder if we shouldn't keep the code in Client as-is and support this env var only for tests. We could rename the env var to PANTHER_ERROR_SCREENSHOT_DIR to be more explicit. WDYT?

Maybe this is the best solution, @nicolas-grekas , wdyt?

nicolas-grekas: Lgtm

dunglas (approved):

nicolas-grekas (approved):

dunglas:
Thanks @kbond

Issue #393, opened 2020-12-09
DomCrawler/Crawler.php throws "Required Param after Optional" Error in PHP8


hawkinbj:
Immediately crashes because Crawler.php constructor looks like this:
![Screen Shot 2020-12-09 at 3 08 48 PM](https://user-images.githubusercontent.com/952976/101695064-49bf1480-3a31-11eb-8aab-6b04078f45c6.png)

While the param $elements is typed, it's default value is [], which causes the error to be thrown. If it were null, no problem. 

More info here: https://php.watch/versions/8.0/deprecate-required-param-after-optional

nicolas-grekas:
Can you please send a PR?

hawkinbj:
https://github.com/hawkinbj/panther/tree/FIX-%23393

PR #394, opened 2020-12-13
[FIX] DomCrawler/Crawler.php throws "Required Param after Optional" Error in PHP8 #393

hawkinbj:
Change empty array default param to null to support PHP8 deprecation of non-null optional params.

### DIFF
--- a/src/DomCrawler/Crawler.php
+++ b/src/DomCrawler/Crawler.php
@@ -34,7 +34,7 @@ final class Crawler extends BaseCrawler implements WebDriverElement
     /**
      * @param WebDriverElement[] $elements
      */
-    public function __construct(array $elements = [], WebDriver $webDriver, ?string $uri = null)
+    public function __construct(array $elements = null, WebDriver $webDriver, ?string $uri = null)
     {
         $this->uri = $uri;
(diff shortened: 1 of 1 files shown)


src/DomCrawler/Crawler.php
```diff
@@ -34,7 +34,7 @@ final class Crawler extends BaseCrawler implements WebDriverElement
     /**
      * @param WebDriverElement[] $elements
      */
-    public function __construct(array $elements = [], WebDriver $webDriver, ?string $uri = null)
+    public function __construct(array $elements = null, WebDriver $webDriver, ?string $uri = null)
```
dunglas, line 37: This doesn't look good to me. It Should be:
```suggestion
    public function __construct(array $elements = [], ?WebDriver $webDriver = null, ?string $uri = null)
```

dunglas (changes requested):

src/DomCrawler/Crawler.php
hawkinbj: See #393

dunglas (approved):

dunglas:
Thank you @hawkinbj

PR #399, opened 2020-12-27
feat: remove ChromeDriver and geckodriver from the repo, use native installations instead

### LINKED ISSUE #83: cannot find Chrome binary
Hy,

on a new installation, I'll testing the example request code on a laravel installation and will become, with the default configuration the following error:

`unknown error: cannot find Chrome binary (Driver info: chromedriver=2.37.544315 (730aa6a5fdba159ac9f4c1e8cbc59bf1b5ce12b7),platform=Linux 4.4.0-131-generic x86_64)`

The binaries are avalaible and executable. I've tested the chromium X64 binary for linux on the command line interface and can resolve the status from the server. 

For the completenes here the code to test `symfony/panther`:

```php
Route::get('test', function(){
    $client = \Symfony\Component\Panther\Client::createChromeClient();
    $crawler = $client->request('GET', 'http://api-platform.com'); // Yes, this website is 100% in JavaScript

    $link = $crawler->selectLink('Support')->link();
    $crawler = $client->click($link);

    // Wait for an element to be rendered
    $client->waitFor('.support');

    echo $crawler->filter('.support')->text();
    $client->takeScreenshot('screen.png');
});
````

### LINKED ISSUE #171: [Question] How to run inside phar
Hello,

I have a small phar, built using humbug/box. It's a symfony/console based application, where I wanted to use panther.
Trying to execute command with Client::createChromeClient() results in:
```
sh: 1: exec: phar:///home/wojciechem/myphar-testing/myphar.phar/vendor/symf  
  ony/panther/src/ProcessManager/../../chromedriver-bin/chromedriver_linux64:  
   not found
```
When I extract the phar contents, the file does exist in expected path.

PHP 7.2.15-1+ubuntu16.04.1+deb.sury.org+1 (cli). I have zip extension installed.

Am I missing something trivial here? Should I use external chromedriver?

### LINKED ISSUE #188: CreatePlatformSocket() returned an error: The requested service provider could not be loaded or initialized. (0x277A) [1556555424.762]
after integrating with laravel I am getting this error is there any way to resolve this problem?
Screenshot (2)

### LINKED ISSUE #262: Chromium not starting in latest version
I am using panther for a while now, and it is working nicely. However since I updated to 0.60.0 chromium is not launching anymore. I am using a docker image php:7-fpm-buster. 

Previously chromium 76 was installed. When running the tests and inspecting the docker container I could see that indeed chromium was started. Since the latest version, using chromium 78, it does not launch at all. I verified that chromium 78 is installed in the container. The odd thing is that I do not get any error or timeout. I let it run for 15 minutes, and no message. Inspecting the container shows that phpunit is running and php is started. 

Running on the latest version of Symfony (4.3.8) in combination with PHPunit 8.3.5.

### LINKED ISSUE #304: Docker issues

### LINKED ISSUE #333: Could not start chrome (or it crashed) after 30 seconds.
Hey! 
Seems like duplicated but I cant solve the issue. 
Starting with firefox driver also causes `Could not start firefox (or it crashed) after 30 seconds.`
Also tries to increase $timeout to 60 secs inside `WebServerReadinessProbeTrait.php`, still no luck.

Server has enough resources 4 cores 16 gb ram.
Composer dependency` "symfony/panther": "^0.7.1"`

Any ideas?

### LINKED ISSUE #356: Install chromedriver matching installed chrome version

### LINKED ISSUE #396: Could not start chrome - error 10106
I am using,

Panther ^0.8.0
Chrome Version 87.0.4280.88 (Official Build) (64-bit)
ChromeDriver 87.0.4280.88
Chrome is installed at C:\Program Files (x86)\Google\Chrome\Application

When I run the code ,
          $client = PantherClient::createChromeClient();
          $crawler = $client->request('GET', 'https://google.com');
          dd($client->getCrawler()->html());

I get,
Unknown error 10106 mapped to net::ERR_FAILED
 [1608463277.815][SEVERE]: CreatePlatformSocket() returned an error: The requested service provider could not be loaded or initialized.

Double clicking on the chromedriver.exe from the panther folder launches a command window with,

Starting ChromeDriver 87.0.4280.88 (89e2380a3e36c3464b5dd1302349b1382549290d-refs/branch-heads/4280@{#1761}) on port 9515
Only local connections are allowed.
Please see https://chromedriver.chromium.org/security-considerations for suggestions on keeping ChromeDriver safe.
ChromeDriver was started successfully.

But nothing more. This is my first time using chromedriver or panther so suspect it is simply a location, version issue but not seeing any logs or anything useful to help diagnose.

If I run a quick python script from the chromedriver folder supplied with panther this works and I see chrome launching and going to the website. 

from selenium import webdriver
from selenium.webdriver.common.keys import Keys
driver = webdriver.Chrome('./chromedriver')
driver.get("https://www.python.org")
print(driver.title)

I've also downloaded the latest laravel dusk package and this runs with no issues. How best to debug the chromedriver installed by panther as I suppose it is related to that

dunglas:
Relying on binaries embedded in the repository has many disadvantages:

1. Drivers shipped in the repo aren't always compatible with local installations of the browsers (outdated, or too recent), causing many issues.
2. Having to update the drivers regularly causes is time-consuming, and shipping outdated versions is annoying for the users.
2. When using GitHub Actions proper installations of ChromeDriver and geckodriver are available by default, relying on them is better. Also, on most OSs it's possible to install ChromeDriver (and often gecokodriver too) using standard package managers (see the updated documentation).
4. Embedded binaries make the repo heavier. This slows down CIs, especially when the drivers aren't necessary, such as when using GitHub Actions.

This PR removes embedded binaries. Panther now relies on the drivers available in the `PATH` instead. If drivers aren't installed, a nice message explaining how to install them locally using @dbrekelmans' Browser Driver Installer is displayed.

I also updated the documentation and fixed some issues in it.

Closes #356, closes #389, closes #358, closes #396, closes #333, closes #304, closes #262, closes #188, closes #171, closes #83.


nicolas-grekas (approved):
Sweet!
Before the tag, what about rewriting the git history to remove those files from it also? We would keep tags as of course.

dunglas:
@nicolas-grekas this will break existing installations. Maybe could we do that at some point (in one year or so), but I think it's too early now.

nicolas-grekas:
If we keep tags, it shouldn't break anything. What do you have in mind?

dunglas:
Yes you’re right, I read too fast, and deleted too slowly 😅. Let’s do this.

dunglas:
I rewrote the Git history of the main branch to remove the binaries (to reduce the size of the repository, as discussed). But it was useless as Git also downloads tags by default and we need to preserve tags to avoid breaking existing apps, so I restored the old history.

This had an undesired side-effect: all open Pull Requests have been close automatically by GitHub. Contributors, sorry about that. Could you rebase and reopen the PR if it is still relevant?

sherazkamran:
**CreatePlatformSocket() returned an error: The requested service provider could not be loaded or initialized. (0x277A).**

Still getting this error. Even after carefully following the documentation. Does anyone know a working solution?

ankitkandari:
@dunglas Is there any solution for this issue...please help

PR #402, opened 2021-01-10
docs: Accessing Browser Console Logs

### LINKED ISSUE #339: Add support to access the browser's console
If you're running Chrome inside a container, it's hard to view its Devtools (haven't found a way yet).

What would be better is to give the developer access to the console log, being able to directly inspect the errors (and later even assert against them).

See
https://github.com/php-webdriver/php-webdriver/issues/570#issuecomment-438640578

dunglas:
Closes #339.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,7 +179,7 @@ class E2eTest extends PantherTestCase
 
 To run this test:
 
-    phpunit tests/E2eTest.php
+    bin/phpunit tests/E2eTest.php
 
 ### A Polymorphic Feline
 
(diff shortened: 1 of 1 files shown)


cookieguru:
This seems to be specific to ChromeDriver; any ideas for geckodriver?

dunglas:
@cookieguru no, GeckoDriver doesn't implement this feature (which isn't part of the W3C WebDriver spec). See https://github.com/mozilla/geckodriver/issues/330.

PR #408, opened 2021-01-27
[Doc] Fix typo

matthieumota:
A little typo with methods on documentation.


dunglas:
Good catch, thanks @matthieumota!

PR #413, opened 2021-02-03
feat: add missing type hints

dunglas:


src/DomCrawler/Field/FormFieldTrait.php
```diff
@@ -34,24 +34,27 @@ public function __construct(WebDriverElement $element)
...
     public function getValue()
     {
         return $this->element->getAttribute('value');
     }
 
-    public function isDisabled()
+    public function isDisabled(): bool
     {
-        return $this->element->getAttribute('disabled') ?? false;
+        return null === $this->element->getAttribute('disabled') ? false : true;
```
nicolas-grekas, line 57: `return null !== $this->element->getAttribute('disabled');`

tests/DummyKernel.php
```diff
@@ -23,11 +23,11 @@
  */
 class DummyKernel implements KernelInterface
 {
-    public function serialize()
+    public function serialize(): void
```
nicolas-grekas, line 26: if that's the serializable interface, it should return `array`

PR #421, opened 2021-02-09
Fixed assertions to work with clients other than the PantherClient

### LINKED ISSUE #420: Call to undefined method when using kernel browser
When you use `assertPageTitleContains` when using a KernelBrowser, then the following exception is thrown:
```
1) App\Tests\AppBundle\Controller\RegistrationControllerTest::testRegisterPageLoads
Error: Call to undefined method Symfony\Bundle\FrameworkBundle\KernelBrowser::createWebDriverByFromLocator()

C:\project_dir\vendor\symfony\panther\src\WebTestAssertionsTrait.php:224
C:\project_dir\vendor\symfony\panther\src\WebTestAssertionsTrait.php:56
C:\project_dir\vendor\symfony\framework-bundle\Test\DomCrawlerAssertionsTrait.php:69
C:\project_dir\vendor\symfony\panther\src\WebTestAssertionsTrait.php:94
C:\project_dir\tests\AppBundle\Controller\RegistrationControllerTest.php:24
```

nusje2000:
Fixes #420

### DIFF
--- a/README.md
+++ b/README.md
@@ -402,6 +402,67 @@ class E2eTest extends PantherTestCase
 }
 ```
 
+### Having a Multi-domain Application
+
+It happens that your PHP/Symfony application might serve several different domain names.
+
+As Panther saves the Client in memory between tests to improve performances, you will have to run your tests in separate
+processes if you write several tests using Panther for different domain names.
+
+To do so, you can use the native `@runInSeparateProcess` PHPUnit annotation.
+
+**ℹ Note:** it is really convenient to use the `external_base_uri` option and start your own webserver in the background,
+because Panther will not have to start and stop your server on each test. [Symfony CLI](https://symfony.com/download) can
+be a quick and easy way to do so. 
+
+Here is an example using the `external_base_uri` option to determine the domain name used by the Client:
+
+```php
+<?php
+
+namespace App\Tests;
+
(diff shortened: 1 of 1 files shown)


src/WebTestAssertionsTrait.php
```diff
@@ -217,10 +227,33 @@ public static function assertSelectorAttributeWillNotContain(string $locator, st
         self::assertSelectorAttributeNotContains($locator, $attribute, $text);
     }
 
+    private static function getText(string $locator): string
+    {
+        $client = self::getClient();
+        if ($client instanceof PantherClient) {
+            return self::findElement($locator)->getText();
+        }
```
nusje2000, line 235: To preserve the current assertion behavior, I used `findElement` when using a `PantherClient` so this is still resolved via the `RemoteWebDriver` instead of using the Crawler. Same goes for `getAttribute`, `assertSelectorExists` and `assertSelectorNotExists`. By doing this, this should not be a breaking change.

dunglas:
I don't know why, but GitHub closed your PR when I pushed to your branch. I reopened it with an additional bug fix and a test in #423.
I've the feeling that we can simplify these assertions by using the Crawler for all clients (Panther and BrowserKit), because under the hood the Panther crawler delegates to BrowserKit. This can be next in a follow-up PR.

PR #430, opened 2021-02-19
Added logging options

### LINKED ISSUE #215: [Feature] Improved logging of managed processes (chrome, webserver)
I've been working on an implementation of panther within a project of mine. This all works fine but resolving errors can be tricky sometimes. I am using panther within a docker container and I had to run the ChromeManager manually to find out that the PHP process didn't have the right permissions to execute. Maybe there could be an option to redirect the output of the chrome manager(and other processes managed by panther) to a separate log file. This feature could make debugging way easier.

### LINKED ISSUE #429: [FEATURE] Add options to enable chrome manager logging
It would be nice that while creating the chrome driver, the options for logging can be passed via the options array. These options can then be used to setup the arguments passed to the chrome manager. The arguments that would be nice to configure are:
 - log-path
 - log-level
 - append-log

nusje2000:
Fixes #215

### DIFF
--- a/src/ProcessManager/ChromeManager.php
+++ b/src/ProcessManager/ChromeManager.php
@@ -110,19 +110,10 @@ final class ChromeManager implements BrowserManagerInterface
 
     private function createProcess(string $chromeDriverBinary): Process
     {
-        $command = [$chromeDriverBinary, '--port='.$this->options['port']];
-
-        if (null !== $this->options['log_path']) {
-            $command[] = '--log-path='.$this->options['log_path'];
-        }
-
-        if (null !== $this->options['log_level']) {
-            $command[] = '--log-level='.$this->options['log_level'];
-        }
-
-        if ($this->options['append_log']) {
-            $command[] = '--append-log';
-        }
+        $command = array_merge(
+            [$chromeDriverBinary, '--port=' . $this->options['port']],
+            $this->options['chromedriver_arguments']
+        );
 
         return new Process($command, null, null, null, null);
     }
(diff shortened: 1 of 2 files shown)


halfer:
Hey @nusje2000, how are you getting on with this? I have the same need as you (and will hack it to get it working in short order).

It looks like the CI is failing for unrelated reasons on 7.1 to 7.3. Is this happening reliably on those PHP versions?

nusje2000:
@halfer I've tested this against PHP 7.2 and that works. The error indeed seems unrelated.

@dunglas would you maybe take a look at this error?

src/ProcessManager/ChromeManager.php
```diff
@@ -108,13 +108,35 @@ private function getDefaultArguments(): array
         return $args;
     }
 
+    private function createProcess(string $chromeDriverBinary): Process
```
dunglas, line 111: Maybe could we simplify this and be more flexible by just creating a `chromedriver_arguments` key allowing to pass an array of random arguments to ChromeDriver?

dunglas:
The error is indeed unrelated, it looks like setup-php now requires to specify explicitly that we need the Zip extension.

src/ProcessManager/ChromeManager.php
nusje2000: I agree, that would be a better solution. I will update this PR later today.

dunglas (approved):

dunglas:
LGTM! Could you also update the readme and the changelog please?

nusje2000:
@dunglas updated the documentation, build should finish now due to the fix made by @halfer.

dunglas:
Thanks @nusje2000

Issue #428, opened 2021-02-18
Adjust the size of the window/screenshot


OskarStark:
I tried sth like:
```php
<?php

declare(strict_types=1);

namespace App\Tests\Acceptance\Util;

use Facebook\WebDriver\WebDriverDimension;
use Symfony\Component\Panther\Client as PantherClient;
use Symfony\Component\Panther\PantherTestCase;

class ScreenshotTestCase extends PantherTestCase
{
    protected static function createPantherClient(array $options = [], array $kernelOptions = [], array $managerOptions = []): PantherClient
    {
        $client = self::createPantherClient();

        $size = new WebDriverDimension(1400, 4000);
        $client->manage()->window()->maximize()->setSize($size);

        return $client;
    }
}
```

But the size of the screenshot doesn't change 🤔

marco-kretz:
Hey there,

had the same problem. Just add `--window-size=1400,900` as chrome arguments :)

shadowc:
Hey, I'm facing the same issue! Could you give us a code example of where chrome arguments go, @marco-kretz ? Thanks!

marco-kretz:
I'm just using env-vars for this. That's what I have in my .env:

`PANTHER_CHROME_ARGUMENTS="--disable-dev-shm-usage --window-size=1400,900 --no-sandbox --headless --hide-scrollbars"`

shadowc:
Ahh, I see! Thank you! I'll try this!

shadowc:
That TOTALLY works! May be in the future we can configure these arguments somewhere within the panther client creation (or a panther.xml.dist file or similar in our projects!)

marco-kretz:
> That TOTALLY works! May be in the future we can configure these arguments somewhere within the panther client creation (or a panther.xml.dist file or similar in our projects!)

You can totally do that already :)

![image](https://user-images.githubusercontent.com/2982363/108548182-c13fd880-72eb-11eb-9c4d-4d0b9e2f800c.png)

shadowc:
I see! May be this is not apparent from `createPantherClient`.

If I do:

```php
$this->client = static::createPantherClient(
    ['external_base_uri' => 'https://localhost:8000'],
    ['HTTPS' => true],
    ['capabilities' => [
        'acceptInsecureCerts' => true,
    ]],
);
```

Because I want to use an external server with https, then I need to figure out how this translates to the `createChromeClient` signature.

shadowc:
From panther's source code:
```php
[...]

PR #437, opened 2021-03-02
createPantherClient improvement

shadowc:
Adds ability to pass browser arguments to the client when a Browser has been specified in the options.

According to what we were able to discuss in #428 , there isn't a clear way to configure a browser window size for the browser (assuming we want to test in interactive mode) from `createPantherClient`.

I commented the following in the issue, looking for answers to this:

```
It would appear by looking at the source code (if I'm not wrong) that if
I call Client::createChromeClient I won't have started a
WebServer (which I guess it's fine in my case, but not ALL cases) and I
won't boot a kernel. What implications does this have for
the test case?
```

Regardless of the implications, I think this is probably a good addition: We can reach firefox/chrome options from `createPantherClient` instead of directly.

Since browser arguments are specific to browser type, I've added a check to nullify this argument IF a browser hasn't been specified by the user. I wonder if this is needed? If I pass a wrong argument to a browser, I suppose the driver will just ignore it and move on.

Let me know if this is a sensible addition or if there is a better way to handle this.

### DIFF
--- a/src/PantherTestCaseTrait.php
+++ b/src/PantherTestCaseTrait.php
@@ -149,11 +149,11 @@ trait PantherTestCaseTrait
      */
     protected static function createPantherClient(array $options = [], array $kernelOptions = [], array $managerOptions = []/*, array $browserArguments = []*/): PantherClient
     {
-        if (\func_num_args() < 4 && __CLASS__ !== (new \ReflectionClass(static::class))->getName() && __CLASS__ !== (new \ReflectionMethod(static::class,__FUNCTION__))->getDeclaringClass()->getName()) {
+        if (\func_num_args() < 4 && __CLASS__ !== (new \ReflectionClass(static::class))->getName() && __CLASS__ !== (new \ReflectionMethod(static::class, __FUNCTION__))->getDeclaringClass()->getName()) {
             @trigger_error(sprintf('The "%s()" method will have a new "array $browserArguments = []" argument in version 1.1.0, not defining it is deprecated since Symfony 1.0.1.', __METHOD__), E_USER_DEPRECATED);
         }
 
-        $browserArguments = 3 < \func_num_args() ? func_get_arg(3) : [];
+        $browserArguments = 3 < \func_num_args() ? \func_get_arg(3) : [];
 
         $browser = ($options['browser'] ?? self::$defaultOptions['browser'] ?? static::CHROME);
         $callGetClient = \is_callable([self::class, 'getClient']) && (new \ReflectionMethod(self::class, 'getClient'))->isStatic();
@@ -169,8 +169,8 @@ trait PantherTestCaseTrait
 
         // Only allow browser arguments if a browser has been explicitly specified
         if (\is_array($browserArguments) && !\array_key_exists('browser', $options)) {
-            @trigger_error('Specifying $browserArguments implies that you specifically set the browser in $options. $browserArguments will be ignored.', E_USER_WARNING);
-            $browserArguments = null;
+            @trigger_error('Specifying $browserArguments requires that you specifically set the browser in $options. $browserArguments will be ignored.', E_USER_WARNING);
+            $browserArguments = [];
         }
 
         self::startWebServer($options);
--- a/src/PantherTestCaseTrait.php
+++ b/src/PantherTestCaseTrait.php
@@ -161,6 +161,7 @@ trait PantherTestCaseTrait
      * Creates the primary browser.
      *
      * @param array $options see {@see $defaultOptions}
+     *
      * @throws \InvalidArgumentException
      */
     protected static function createPantherClient(array $options = [], array $kernelOptions = [], array $managerOptions = []): PantherClient
@@ -186,9 +187,9 @@ trait PantherTestCaseTrait
         if (\array_key_exists('browser_arguments', $options)) {
             if (!\is_array($options['browser_arguments'])) {
                 throw new \InvalidArgumentException('Expected key "browser_arguments" to be an array.');
-            } else {
-                $browserArguments = $options['browser_arguments'];
             }
+
+            $browserArguments = $options['browser_arguments'];
         }
 
         if (PantherTestCase::CHROME === $browser) {
--- a/src/PantherTestCaseTrait.php
+++ b/src/PantherTestCaseTrait.php
@@ -149,8 +149,8 @@ trait PantherTestCaseTrait
      */
     protected static function createPantherClient(array $options = [], array $kernelOptions = [], array $managerOptions = []/*, array $browserArguments = []*/): PantherClient
     {
-        if (\func_num_args() < 4 && \__CLASS__ !== (new \ReflectionClass(static::class))->getName() && \__CLASS__ !== (new \ReflectionMethod(static::class, \__FUNCTION__))->getDeclaringClass()->getName()) {
-            @trigger_error(sprintf('The "%s()" method will have a new "array $browserArguments = []" argument in version 1.1.0, not defining it is deprecated since Symfony 1.0.1.', \__METHOD__), \E_USER_DEPRECATED);
+        if (\func_num_args() < 4 && __CLASS__ !== (new \ReflectionClass(static::class))->getName() && __CLASS__ !== (new \ReflectionMethod(static::class, __FUNCTION__))->getDeclaringClass()->getName()) {
+            @trigger_error(sprintf('The "%s()" method will have a new "array $browserArguments = []" argument in version 1.1.0, not defining it is deprecated since Symfony 1.0.1.', __METHOD__), \E_USER_DEPRECATED);
         }
 
         $browserArguments = 3 < \func_num_args() ? \func_get_arg(3) : [];
--- a/src/PantherTestCaseTrait.php
+++ b/src/PantherTestCaseTrait.php
@@ -147,7 +147,7 @@ trait PantherTestCaseTrait
      *
      * @param array $options see {@see $defaultOptions}
      */
-    protected static function createPantherClient(array $options = [], array $kernelOptions = [], array $managerOptions = []): PantherClient
+    protected static function createPantherClient(array $options = [], array $kernelOptions = [], array $managerOptions = [], array $browserArguments = null): PantherClient
     {
         $browser = ($options['browser'] ?? self::$defaultOptions['browser'] ?? static::CHROME);
         $callGetClient = \is_callable([self::class, 'getClient']) && (new \ReflectionMethod(self::class, 'getClient'))->isStatic();
@@ -161,12 +161,17 @@ trait PantherTestCaseTrait
             }
         }
 
+        // Only allow browser arguments if a browser has been explicitly specified
+        if (is_array($browserArguments) && !array_key_exists('browser', $options)) {
+            $browserArguments = null;
+        }
+
         self::startWebServer($options);
 
         if (static::CHROME === $browser) {
-            self::$pantherClients[0] = self::$pantherClient = Client::createChromeClient(null, null, $managerOptions, self::$baseUri);
+            self::$pantherClients[0] = self::$pantherClient = Client::createChromeClient(null, $browserArguments, $managerOptions, self::$baseUri);
         } else {
-            self::$pantherClients[0] = self::$pantherClient = Client::createFirefoxClient(null, null, $managerOptions, self::$baseUri);
+            self::$pantherClients[0] = self::$pantherClient = Client::createFirefoxClient(null, $browserArguments, $managerOptions, self::$baseUri);
         }
 
         if (\is_a(self::class, KernelTestCase::class, true)) {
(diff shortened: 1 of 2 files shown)


nusje2000 (approved):

shadowc:
Fixes #428

dunglas:
Isn't this change a BC break?

nusje2000:
@dunglas since the new argument is optional and is introduced at the end of the argument list, it should not break anything.

dunglas:
It will. This is a protected method that can be overrode. https://symfony.com/doc/current/contributing/code/bc.html#changing-classes

shadowc:
Will fix soon!

src/PantherTestCaseTrait.php
```diff
@@ -147,7 +147,7 @@ public static function isWebServerStarted(): bool
      *
      * @param array $options see {@see $defaultOptions}
      */
-    protected static function createPantherClient(array $options = [], array $kernelOptions = [], array $managerOptions = []): PantherClient
+    protected static function createPantherClient(array $options = [], array $kernelOptions = [], array $managerOptions = [], array $browserArguments = null): PantherClient
```
OskarStark, line 150:
```suggestion
    protected static function createPantherClient(array $options = [], array $kernelOptions = [], array $managerOptions = [], array $browserArguments = []): PantherClient
```
??

src/PantherTestCaseTrait.php
```diff
@@ -161,12 +161,17 @@ protected static function createPantherClient(array $options = [], array $kernel
             }
         }
 
+        // Only allow browser arguments if a browser has been explicitly specified
+        if (\is_array($browserArguments) && !\array_key_exists('browser', $options)) {
```
OskarStark, line 165: If no browser is specified and browser arguments are set this should result in an exception to let the user know, these browser arguments will not be taken into account, WDYT?

shadowc: I agree, we might raise an exception. Although, I'm not sure of the right wording or exception subclass.

src/PantherTestCaseTrait.php
weaverryan: yes, passing an empty array will not break anything

shadowc: Done

src/PantherTestCaseTrait.php
shadowc: I'm triggering a warning now.

shadowc:
It looks like tests are failing on Ubuntu becuase Chrome Driver crashes, but my local tests in Ubuntu 20.04 with PHP 7.4 are passing.

src/PantherTestCaseTrait.php
```diff
@@ -147,8 +147,14 @@ public static function isWebServerStarted(): bool
      *
      * @param array $options see {@see $defaultOptions}
      */
-    protected static function createPantherClient(array $options = [], array $kernelOptions = [], array $managerOptions = []): PantherClient
+    protected static function createPantherClient(array $options = [], array $kernelOptions = [], array $managerOptions = []/*, array $browserArguments = []*/): PantherClient
     {
+        if (\func_num_args() < 4 && __CLASS__ !== (new \ReflectionClass(static::class))->getName() && __CLASS__ !== (new \ReflectionMethod(static::class, __FUNCTION__))->getDeclaringClass()->getName()) {
```
shadowc, line 152: I've adapted this from this example:

```php
         if (\func_num_args() < 3 && __CLASS__ !== \get_class($this) && __CLASS__ !== (new \ReflectionMethod($this, __FUNCTION__))->getDeclaringClass()->getName() && !$this instanceof \PHPUnit\Framework\MockObject\MockObject && !$this instanceof \Prophecy\Prophecy\ProphecySubjectInterface) {
```
But I'm not sure if I'm doing the right thing here. May be a test is in order.

src/PantherTestCaseTrait.php
```diff
@@ -147,8 +147,14 @@ public static function isWebServerStarted(): bool
...
+            @trigger_error(sprintf('The "%s()" method will have a new "array $browserArguments = []" argument in version 1.1.0, not defining it is deprecated since Symfony 1.0.1.', __METHOD__), \E_USER_DEPRECATED);
```
shadowc, line 153: I'm totally making up the version numbers in this message. Somebody should make them real. :)

src/PantherTestCaseTrait.php
```diff
@@ -161,12 +167,18 @@ protected static function createPantherClient(array $options = [], array $kernel
             }
         }
 
+        // Only allow browser arguments if a browser has been explicitly specified
+        if (\is_array($browserArguments) && !\array_key_exists('browser', $options)) {
+            @trigger_error('Specifying $browserArguments requires that you specifically set the browser in $options. $browserArguments will be ignored.', \E_USER_WARNING);
```
shadowc, line 172: I'm now triggering this warning if a user failed to specify a browser while adding browser arguments.

src/PantherTestCaseTrait.php
```diff
@@ -161,12 +167,18 @@ protected static function createPantherClient(array $options = [], array $kernel
...
+            $browserArguments = [];
+        }
+
         self::startWebServer($options);
 
         if (static::CHROME === $browser) {
-            self::$pantherClients[0] = self::$pantherClient = Client::createChromeClient(null, null, $managerOptions, self::$baseUri);
+            self::$pantherClients[0] = self::$pantherClient = Client::createChromeClient(null, $browserArguments, $managerOptions, self::$baseUri);
```
nusje2000, line 179: Can't we use the provided $options array to add an additional key for "browser_arguments", then this change won't be breaking.

src/PantherTestCaseTrait.php
```diff
@@ -161,12 +167,18 @@ protected static function createPantherClient(array $options = [], array $kernel
             }
         }
 
+        // Only allow browser arguments if a browser has been explicitly specified
```
nusje2000, line 170: Why would you have to explicitly set the used browser to provide browser arguments ?

shadowc: This is a question I've asked myself. I'm not sure if passing a wrong argument would cause any issues. May be this shouldn't be enforced at all.

nusje2000: I think users should be able to rely on chrome being the default, so providing arguments without explicitly setting the browser should be allowed.

src/PantherTestCaseTrait.php
shadowc: Hmm, I think I like this idea much better. Simpler, no deprecation warnings, we just transfer `$options['browser_arguments']` into that third argument and everything just works(tm)

src/PantherTestCaseTrait.php
shadowc: Ok, I agree! I'm not expert on browser drivers, so I just wanted to get users into a hypotetical crash due to passing wrong arguments. I don't think this would ever happen, so I can remove this altogether.

shadowc:
I've pushed the approach proposed by @nusje2000 . What do you all think?

src/PantherTestCaseTrait.php
```diff
@@ -163,10 +163,12 @@ protected static function createPantherClient(array $options = [], array $kernel
 
         self::startWebServer($options);
 
+        $browserArguments = \array_key_exists('browser_arguments', $options) && \is_array($options['browser_arguments']) ? $options['browser_arguments'] : [];
```
nusje2000, line 166: I think a possible interpretation of this key might be that you can put a string of options in this like `--some-option --some-other-option`. This of course should not be supported, but I think this mistake can be easily made. Should we raise an exception if this is not an array and just say it must be an array.

With the current implementation, incorrect usage is suppressed. We might be better of forcing correct usage.

shadowc:
Any idea why tests are failing left right and center with this error message?

```
) Symfony\Component\Panther\Tests\AssertionsTest::testDomCrawlerAssertions with data set "Panther" (array('Symfony\Component\Panther\Tes...nsTest', 'createPantherClient'), 'Symfony\Component\Panther\Client')
Facebook\WebDriver\Exception\UnknownErrorException: unknown error: Chrome failed to start: exited abnormally.
  (unknown error: DevToolsActivePort file doesn't exist)
  (The process started from chrome location /usr/bin/google-chrome is no longer running, so ChromeDriver is assuming that Chrome has crashed.)
```

src/PantherTestCaseTrait.php
shadowc: Do you mean just make sure that `$options['browser_options']` is not a string?

nusje2000: I think the best option is to make sure that `browser_arguments` is an array so adding this should be sufficient:
```php
if (!is_array($browserArguments )) {
    throw new UnexpectedValueException('Expected key "browser_arguments" to be an array.');
}

Comment (nusje2000) [inline src/PantherTestCaseTrait.php:166]: The default value should be null since an empty array would remove all the default arguments.

Comment (shadowc) [reply src/PantherTestCaseTrait.php:166]: Ahh, ok! Will switch default to `null` again.

Comment (shadowc) [reply src/PantherTestCaseTrait.php:166]: I think `!\is_array` is what I'm thinking about

Comment (shadowc) [reply src/PantherTestCaseTrait.php:166]: @nusje2000 how about throwing a warning using

```php
@trigger_error('...', \E_USER_WARNING);
```

nusje2000 (approved):

shadowc:
Alright. One last change. Defaulting `browser_arguments` to `null` made all the tests pass! Now also triggering a warning when `$options['browser_arguments']` is not an array!

nusje2000 (approved):

src/PantherTestCaseTrait.php
weaverryan: It's always the next MAJOR version where the BC break would occur. So this should be 2.0 instead of 1.1

src/PantherTestCaseTrait.php
```diff
@@ -163,10 +163,20 @@ protected static function createPantherClient(array $options = [], array $kernel
 
         self::startWebServer($options);
 
+        $browserArguments = null;
+
+        if (\array_key_exists('browser_arguments', $options)) {
+            if (!\is_array($options['browser_arguments'])) {
+                @trigger_error('Expected key "browser_arguments" to be an array.', \E_USER_WARNING);
```
dunglas, line 170: You should throw an `\InvalidArgumentException` instead. We don't trigger errors (except for deprecation notices) in Symfony.

dunglas (changes requested):
Could you also add a test please?

dunglas:
Hi @shadowc! Gentle ping :) Do you still have the time to work on this PR or do you want someone else to try to finish it? Thanks!

shadowc:
I'm so sorry for disappearing! There have been many IRL events that prevented me from doing ANY side work. I WANT to get back to this this weekend if that's ok with you!

dunglas:
No worries, it's perfectly OK to have more important things to do. We aren't paid for this, after all!

shadowc:
Hey, @dunglas ! So now I've corrected the file (also fixed conflicts with latest `main`) and wrote some tests, although I'm not sure if those are done in a correct way.

I am getting a timeout on PHP 8.0 on Mac which I'm not 100% sure it's unstable or a mistake on the way I'm deleting the client from one test to the next.

shadowc:
Ping @dunglas ! So my tests are failing due to something I cannot really understand (looks like an artifact on the github action? I don't have a Mac unfortunately). Let me know if I can do something this weekend to help with it. (Also my tests might not be what you want/expect, so you can give me some feedback on that as well) :)

shadowc:
Should I Ping again? I assume this PR is still relevant?

tests/ClientTest.php
```diff
@@ -25,13 +25,41 @@
...
+        $this->assertInstanceOf(AbstractBrowser::class, $client);
+        $this->assertInstanceOf(WebDriver::class, $client);
+        $this->assertInstanceOf(JavaScriptExecutor::class, $client);
+        $this->assertInstanceOf(KernelInterface::class, self::$kernel);
+
+        // I'm not sure exactly how to start with a clear class, since the
+        // test file assumes the client will be present for the next tests
+        // So I'm cleaning up this way here so that we can have a default
+        // client next
+        self::$pantherClient = null;
```
weaverryan, line 60: This is tricky. You're re-setting this so that `createPantherClient()` won't reuse your instance in further tests. However, it has a side effect that the `stopWebServer()` in the trait (called by `tearDownAfterClass()`) will not fire correctly.

Try `$this->stopWebServer()` instead:

```php
// stop the web server and clean up self::$pantherClient so it is not reused by other tests
$this->stopWebServer();
```

shadowc: Trying with this change now!

shadowc: wow, tests are passing now! That is great! :)

weaverryan (approved):

shadowc:
This PR is green and ready for review.

YasinAshorian (approved):
👍

PR #438, opened 2021-03-02
Undefined class constant 'CHROME' fix

### LINKED ISSUE #384: Undefined class constant 'CHROME'
Hi,

Symfony 4.4 fails when using `PantherTestCaseTrait`:

> Undefined class constant 'CHROME'

I see https://github.com/symfony/panther/issues/317 and that it is merged but the error is still happening (starting with 0.7.0 and to the latest).
I created a reproducible example https://github.com/aurelijusrozenas/panther-test

The test code taken from example and is minimal: https://github.com/aurelijusrozenas/panther-test/blob/main/tests/DefaultTest.php

Any ideas? I would say it's a bug but how come no one has reported yet? :)

nusje2000:
fixes #384

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,6 +312,35 @@ class ConsoleTest extends PantherTestCase
 }
 ```
 
+### Configuring Chrome Manager Options
+
+If needed, you can configure arguments passed to the chrome manager binary:
+
+```php
+<?php
+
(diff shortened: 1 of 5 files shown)


nusje2000:
Found out that there where still unrelated changes in my fork, so I declined this PR.

PR #439, opened 2021-03-02
Fixed undefined constant error when using PantherTestCaseTrait

### LINKED ISSUE #384: Undefined class constant 'CHROME'

nusje2000:
Fixes #384


OskarStark (approved):

dunglas:
I prefer to wait for native support of enums that will land in PHP 8.1 instead of doing our own thing.

src/Browser.php
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Symfony\Component\Panther;
+
+final class Browser
```
jderusse, line 16: `Browers` for consistency with sulfony's `Events`?

nusje2000: I agree, I will fix this if we indeed want this PR to go to master.

nusje2000:
@dunglas the transition from this class to an enum is probably not that complex and since symfony now supports 7.1, we would probably talking about years in the future for this to be possible. The current implementation is very simple and is only for moving the responsibility from the TestCase to a seperate class.

Another option imo would be to move the constants to the trait instead of the PantherTestCase, this would make more sense since now the trait does not have a dependency on its parent.

What do you think?

nusje2000:
@dunglas  would you mind to have a look at this PR again, I changed the scope to only fix the bug without introducing a new enum.

dunglas (approved):

nusje2000:
@dunglas can this PR get merged?

dunglas:
Thank you @nusje2000!

Issue #427, opened 2021-02-14
Browser closes on interactive mode after test fail


shadowc:
Hey, either I'm doing something wrong or there's something going on with interactive mode.

phpunit.xml:

```
<?xml version="1.0" encoding="UTF-8"?>
<!-- https://phpunit.readthedocs.io/en/latest/configuration.html -->
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.3/phpunit.xsd" backupGlobals="false" colors="true" bootstrap="tests/bootstrap.php">
  <coverage processUncoveredFiles="true">
    <include>
      <directory suffix=".php">src</directory>
    </include>
  </coverage>

  <php>
    <ini name="error_reporting" value="-1"/>
    <server name="APP_ENV" value="test" force="true"/>
    <server name="SHELL_VERBOSITY" value="-1"/>
    <server name="SYMFONY_PHPUNIT_REMOVE" value=""/>
    <server name="SYMFONY_PHPUNIT_VERSION" value="9.5.1"/>
  </php>

  <testsuites>
    <testsuite name="Project Test Suite">
      <directory>tests</directory>
    </testsuite>
  </testsuites>

  <listeners>
    <listener class="Symfony\Bridge\PhpUnit\SymfonyTestsListener"/>
  </listeners>

  <extensions>
    <extension class="DAMA\DoctrineTestBundle\PHPUnit\PHPUnitExtension"/>
    <extension class="Symfony\Component\Panther\ServerExtension"/>
  </extensions>
</phpunit>
```

command:
```
PANTHER_NO_HEADLESS=1 bin/phpunit --debug
```

console output:

```
ime: 00:09.015, Memory: 50.50 MB

There was 1 failure:

1) App\Tests\e2e\AdminBasicTest::testBasicAdminPageLoads
Fail to find the logout link

/home/shadowc/web-local/decoramemas/tests/e2e/AdminBasicTest.php:37

FAILURES!
Tests: 142, Assertions: 622, Failures: 1.

Remaining indirect deprecation notices (1)

  1x: The "DAMA\DoctrineTestBundle\Doctrine\DBAL\AbstractStaticDriverV2" class implements "Doctrine\DBAL\Driver\ExceptionConverterDriver" that is deprecated.
    1x in PHPUnitExtension::executeBeforeFirstTest from DAMA\DoctrineTestBundle\PHPUnit

[...]

PR #440, opened 2021-03-07
ServerTrait: Don't check isWebServerStarted on pause

shadowc:
For the long story, check #427 .

After some investigation, I've found out that we aren't really testing weather a browser window stays open after a test failure or error (this might be untestable, tbf):

(From the issue posts)

------

I've found the test that is supposed to check weather the browser will be paused on test failure. The problem with the test itself is that it skips calling `$this->pause(sprintf('Failure: %s', $message));`:

```php
public function testPauseOnFailure(string $method, string $expected): void
{
    $extension = new ServerExtension();
    $extension->testing = true;
    ...
```

In the beginning of the test, the flag `$extension->testing` is set. Later, in `pause()` we see this:

```php
if (!$this->testing) {
    fgets(\STDIN);
}
```

And this is why a test wouldn't cover the issue reported here. So I added some `echo` lines throughout the code to debug a little bit and I've found out that in my particular case, when I reach the test failure, `PantherTestCase::isWebServerStarted()` is `false`, so the pause is never done:

(with my debug lines)
```php
private function pause($message): void
{
    echo PantherTestCase::isWebServerStarted() ? "Web server started\n" : "Web server hasn't started!!\n";

    if (PantherTestCase::isWebServerStarted()
        && \in_array('--debug', $_SERVER['argv'], true)
        && $_SERVER['PANTHER_NO_HEADLESS'] ?? false
    ) {
        echo "$message\n\nPress enter to continue...";
        if (!$this->testing) {
            fgets(\STDIN);
        }
    }
}
```

And the output is:

```
Testing /home/shadowc/web-local/decoramemas/tests/e2e
Test 'App\Tests\e2e\AdminBasicTest::testBasicAdminPageLoads' started
Web server hasn't started!!
Test 'App\Tests\e2e\AdminBasicTest::testBasicAdminPageLoads' ended
```

Thus the pause isn't performed.

-----------

By just removing that check in `pause` I was able to make the browser window stick after a test error or failure.

### DIFF
--- a/src/ServerTrait.php
+++ b/src/ServerTrait.php
@@ -34,8 +34,7 @@ trait ServerTrait
 
     private function pause($message): void
     {
-        if (PantherTestCase::isWebServerStarted()
-            && \in_array('--debug', $_SERVER['argv'], true)
+        if (\in_array('--debug', $_SERVER['argv'], true)
             && $_SERVER['PANTHER_NO_HEADLESS'] ?? false
         ) {
             echo "$message\n\nPress enter to continue...";


weaverryan:
@shadowc if you remove this line locally and test, once it hits pause, is the "Panther web server" actually running? What I mean is, *are* you able to actually browse the site? If you are, then the Panther web server IS running... but for some reason, this `PantherTestCase::isWebServerStarted()` is returning `false` incorrectly. If it is NOT running, then this means that Panther has incorrectly stopped its web server too early (and even if we removed this line, the user wouldn't *actually* be able to surf their site).

Cheers!

shadowc:
After some discussions, we came to the conclusion that the reason why `PantherTestCase::isWebServerStarted()` is returning `false` for me is because I'm using a custom server to run panther in.

Therefore the question for this PR should be:

- Do we care if panther webserver is running before pause?
- Or do we add an exception to pause anyways if the user is running their custom web server?

Also, Ryan suggested something along these lines:

```php
private function pause($message): void
{
    if (PantherTestCase::isWebServerStarted()
        && \in_array('--debug', $_SERVER['argv'], true)
        && $_SERVER['PANTHER_NO_HEADLESS'] ?? false
    ) {
        echo "$message\n\nPress enter to continue...";
        if (!$this->testing) {
            fgets(\STDIN);
        }
    } else {
        echo "$message\n\nCould not pause: web server is not running..."
    }
}
```

Where we put an `else` clause so that we can output something if Panther is skipping the pause. This way, the user has more information as to what is happening.

I'm leaning more towards adding the exception in the form `PantherTestCase::isWebServerStarted() || <user is using custom server>` to the `if` statement.

weaverryan:
> I'm leaning more towards adding the exception in the form PantherTestCase::isWebServerStarted() || <user is using custom server> to the if statement.

I agree. I can't think of a reason why pausing when the user is using a "custom server" would be a problem at all. Are we easily able to add this check here? Do we have a way (from this trait method) to be able to check if the user is using a custom server?

shadowc:
I'd need to dig into the code to see what's available in the form of data. This is probably available as a static member of `PantherTestCase`, or can be made available.

ThomasTr:
I have the same problem: with PANTHER_NO_HEADLESS=1 and --debug my browser window does not stay open when testcase fails. With this change my browser window stays open in said case.

nclavaud:
Same issue here. I guess we are left with two options:

1. Remove the condition

```diff
// src/ServerTrait.php

- if (PantherTestCase::isWebServerStarted()
-      && \in_array('--debug', $_SERVER['argv'], true)
+ if (\in_array('--debug', $_SERVER['argv'], true)
      && $_SERVER['PANTHER_NO_HEADLESS'] ?? false
  ) {
      // ...
  }
```

or

2. Be more specific

```diff
// src/ServerTrait.php

- if (PantherTestCase::isWebServerStarted()
+ if ((!PantherTestCase::isWebServerManaged() || PantherTestCase::isWebServerStarted())
      && \in_array('--debug', $_SERVER['argv'], true)
      && $_SERVER['PANTHER_NO_HEADLESS'] ?? false
  ) {
      // ...
  }

// src/PantherTestCaseTrait.php

+ public static function isWebServerManaged(): bool
+ {
+     return null !== self::$webServerManager;
+ }
```

dunglas:
Merging as is for now to fix the bug, but I like your second proposal @nclavaud. Feel free to open a PR!

dunglas:
Thank you @shadowc! And sorry for the long delay before the merge.

shadowc:
Thank you guys! I've been completely out of this project due to many many IRL stuff happening. I WANT to get back to it (other PR) this weekend though!

PR #444, opened 2021-03-14
Allow options chromeManager

Damienbelingheri:
Allows the possibilities to pass options to the ChromeManager.
We can't do any test because the passed capabilities are discarded by the RemoveWebDriver.


Damienbelingheri:
Test doesn't pass because of a weird timeout, but the code work.

dunglas (approved):

dunglas:
Thanks @Damienbelingheri

PR #451, opened 2021-04-08
Allow custom environment variables

ThomasTr:
Make the sixth parameter ($env) of the WebServerManager class configurable. 
Very useful to call e.g. different test_case.


dunglas:
Thanks @ThomasTr. I renamed the option `env` for consistency with the parameter name.

PR #455, opened 2021-04-14
Minor rewordings

ThomasLandauer:


README.md
```diff
@@ -45,8 +45,6 @@ Use [Composer](https://getcomposer.org/) to install Panther in your project. You
 
 ### Installing ChromeDriver and geckodriver
 
-Panther uses the WebDriver protocol to control the browser used to crawl websites.
```
dunglas, line 48: Why should we delete this? It's an important information.

README.md
```diff
@@ -68,9 +66,9 @@ On Windows, using [chocolatey](https://chocolatey.org):
 
     choco install chromedriver selenium-gecko-driver
 
-Finally, you can download manually [ChromeDriver](https://sites.google.com/a/chromium.org/chromedriver/) (for Chromium or Chrome)
-and [GeckoDriver](https://github.com/mozilla/geckodriver) (for Firefox) and put them anywhere in your `PATH`
-or in the `drivers/` directory of your project.
+Finally, you can manually download [ChromeDriver](https://sites.google.com/a/chromium.org/chromedriver/) (for Chromium or Chrome)
+or [GeckoDriver](https://github.com/mozilla/geckodriver) (for Firefox) and add them to your `PATH` or put them
```
dunglas, line 70: `and` is better (you may need both).

dunglas (changes requested):

README.md
ThomasLandauer: A quick Ctrl+F for "webdriver" gives me:
* It leverages the W3C's WebDriver protocol
* $client->getWebDriver()->manage()->getLog('browser');
* client->getWebDriver()->manage()->getLog('performance');
* Checking the State of the WebDriver Connection
* Use the Client::ping() method to check if the WebDriver connection is still active
* because this library uses WebDriverElement internally
* Panther is built on top of PHP WebDriver

So if you merge this, it will still be explained 7 times :-)

README.md
ThomasLandauer: I changed it to "and/or", since not everybody needs both. OK?

README.md
dunglas: It's stated only one time that it's a W3C spec and it's an important information (especially because there are several things named WebDriver).

ThomasLandauer: Sorry, I don't see your point.

> It's stated only one time that it's a W3C spec

Well, but the sentence I deleted doesn't contain that! ;-)
Besides, the "installation" section just isn't the right place to inform people what this project is all about. This should be explained at the very top, and the third sentence on the page does this very clear IMO:

> It leverages the W3C's WebDriver protocol to drive native web browsers such as Google Chrome and Firefox.

PR #456, opened 2021-04-15
Adding self-signed certificates for Firefox

### LINKED ISSUE #335: Panther fails on Curl certificate validation on readiness trait
Hi

It doesn't seem possible to use Panther without valid ssl certs. Whilst the environment variable switch can tell chrome to ignore certificate errors, `WebServerReadinessTrait::waitUntilReady` generates a curl client that is fails due to invalid certificates. 

Switching out

`$client = HttpClient::create(['timeout' => $timeout]);`

for

`$client = HttpClient::create(['timeout' => $timeout, 'verify_host' => false, 'verify_peer' => false]);`

enables me to test with Panther but this doesn't seem achievable via config, or am I missing something?

Ben

ThomasLandauer:
Closes #335


dunglas:
Thanks @ThomasLandauer

PR #463, opened 2021-04-16
Explaining "visible" as within the scrollarea

ThomasLandauer:
Maybe this information could be merged with "Accessing To Hidden Text" into a new heading?

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,8 @@ echo $crawler->filter('#installing-the-framework')->text();
 $client->takeScreenshot('screen.png'); // Yeah, screenshot!
 ```
 
+"visible" means that the text is within the current scroll area. To check content further down on the page, you can adjust the window size (see below).
+
 ### Testing Usage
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -120,6 +120,8 @@ echo $crawler->filter('#installing-the-framework')->text();
 $client->takeScreenshot('screen.png'); // Yeah, screenshot!
 ```
 
+"visible" means that the text is within the current scroll area. To check content further down on the page, you can adjust the window size (see below).
```
dunglas, line 123: This is actually a bit more complex than that: https://www.w3.org/TR/webdriver/#element-displayedness

Maybe can we just point to the spec?

dunglas (changes requested):

README.md
ThomasLandauer: Sorry, I don't understand what their bottom line is :-(
Please decide yourself. My point was to give some explanation (or hint at least) if "visible" just means "present in the source code" or "actually displayed on the current screen".

PR #465, opened 2021-04-19
Adding Port

ThomasLandauer:
Haven't tested for Chrome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -388,6 +388,15 @@ $client = self::createPantherClient([
 * `PANTHER_FIREFOX_ARGUMENTS`: to customize Firefox arguments. You need to set `PANTHER_NO_HEADLESS` to fully customize.
 * `PANTHER_FIREFOX_BINARY`: to use another `firefox` binary
 
+### Changing the Port
+
+By default, chromedriver is listening on port 9515, and geckodriver on 4444. To change this value, pass another port to the `$options` array:
+
+```php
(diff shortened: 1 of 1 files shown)


dunglas:
I would prefer adding a more generic section explaining how to pass options to the drivers, similar to the existing `Passing Arguments to ChromeDriver` section. WDYT?

ThomasLandauer:
No, sorry, I have too little overview.
* Some headings explicitly mention Chrome (e.g. [Passing Arguments to ChromeDriver](https://github.com/symfony/panther#passing-arguments-to-chromedriver)
* Some headings mention a topic (but no platform), but the text only addresses Chrome [Accepting Self-signed SSL Certificates](https://github.com/symfony/panther#accepting-self-signed-ssl-certificates), see https://github.com/symfony/panther/pull/456

So while I do agree with what I think is your bottom line (namely that README is a bit unorganized right now :-) I can't guess in which direction you want to evolve it ;-)
Ultimately, the page is getting to long anyway, and should be separated into a dedicated (sub)website. But since this would probably be a larger undertaking, the best guess for now is probably: Merge this as is, and do the larger reorganization later...

dunglas:
Maybe should we just move these docs in `symfony/symfony-docs`? Panther reached 1.0 anyway, so it's probably the good time. WDYT @symfony/team-symfony-docs?

PR #468, opened 2021-04-27
Option for disable devtools

### LINKED ISSUE #168: Dev-tool always open
Hello !

when i launch the test running PANTHER_NO_HEADLESS=1 ./bin/phpunit on my symfony 4.2 project, i can see than chrome open with the dev tool, and this makes my tests not working well because of my bootstrap 4 navbar. Any possibility to configure it ?

kacpermajczak:
Fixes #168

Changes are backwards compatible.

For disabling dev tools just add `PANTHER_DEVTOOLS=0` before the command or add it as a global server variable
```
PANTHER_DEVTOOLS=0 php bin/phpunit  
```


dunglas:
Thanks @kacpermajczak

PR #474, opened 2021-05-23
Add option to attach screenshots to phpunit junit report

trakos:
There is a fun little feature in gitlab, that allows to link to attached screenshot when showing test failure details:

image

After clicking link:

image

The screenshot has to be linked to in junit file, in the `<system-out>` element in a format `[[ATTACHMENT|./relative-url.png]]`, as described here: https://docs.gitlab.com/ee/ci/unit_test_reports.html#viewing-junit-screenshots-on-gitlab (note: for now documentation states that the path has to be absolute, but it's not true - there's an issue about it https://gitlab.com/gitlab-org/gitlab/-/issues/330951, I've tested it and the path has to be relative to build directory).

Example `<testcase>` entry in junit from gitlab documentation:

```
<testcase time="1.00" name="Test">
  <system-out>[[ATTACHMENT|/absolute/path/to/some/file]]</system-out>
</testcase>
```

It can actually easily be achieved by echoing this string during the test, phpunit picks that up when using junit log format. I'm proposing adding simple option `PANTHER_ERROR_SCREENSHOT_ATTACH` that would print such string, to allow using this with gitlab, and possibly with jenkins plugin https://plugins.jenkins.io/junit-attachments/ .

I've tried to get better support from phpunit for this attachment format, but it got nowhere: https://github.com/sebastianbergmann/phpunit/pull/4666

The only problem was that the screenshot cannot be taken in `executeAfterTestFailure`, as phpunit no longer treats output as test output then. I have moved screenshot taking to `tearDown`, and made `ServerExtension::takeScreenshots` public.

Would you consider adding such option to panther?

### DIFF
--- a/src/ServerExtension.php
+++ b/src/ServerExtension.php
@@ -63,13 +63,11 @@ final class ServerExtension implements BeforeFirstTestHook, AfterLastTestHook, B
 
     public function executeAfterTestError(string $test, string $message, float $time): void
     {
-        $this->takeScreenshots('error', $test);
         $this->pause(sprintf('Error: %s', $message));
     }
 
     public function executeAfterTestFailure(string $test, string $message, float $time): void
     {
-        $this->takeScreenshots('failure', $test);
         $this->pause(sprintf('Failure: %s', $message));
     }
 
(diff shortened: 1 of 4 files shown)


src/ServerExtension.php
```diff
@@ -78,20 +76,24 @@ private static function reset(): void
...
+            $screenshotPath = sprintf('%s/%s_%s_%s-%d.png',
                 $_SERVER['PANTHER_ERROR_SCREENSHOT_DIR'],
                 date('Y-m-d_H-i-s'),
                 $type,
                 strtr($test, ['\\' => '-', ':' => '_']),
                 $i
-            ));
+            );
+            $client->takeScreenshot($screenshotPath);
+            if ($_SERVER['PANTHER_ERROR_SCREENSHOT_ATTACH'] ?? false) {
```
dunglas, line 94: Is there a drawback to always call `printf`?

trakos: PHPunit has `--disallow-test-output` option on the command line and `beStrictAboutOutputDuringTests="true"` setting in xml. Additionally, there's an `expectOutputString` method in phpunit (though I doubt someone would use it with panther).

dunglas (approved):

dunglas:
Thanks @trakos

dmitryuk:
Looks like the functional doesn't work more with PhpUnit > 10

PR #509, opened 2021-10-30
Add support for Symfony 6

jordisala1991:
It is required to run with dev dependencies, otherwise it wont be installed. I didn't found a dev dependencies build so I added one (if it is not correct this way, I can revert it or change to another one)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
             - name: Setup PHP
               uses: shivammathur/setup-php@v2
               with:
-                  php-version: '7.4' # PHP CS Fixer isn't compatible with PHP 8 yet https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4702
+                  php-version: '8.0'
                   tools: php-cs-fixer, cs2pr
 
             - name: PHP Coding Standards Fixer
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,6 +88,43 @@ jobs:
             - name: Run tests
               run: vendor/bin/simple-phpunit
 
+    phpunit-dev:
+        runs-on: ubuntu-latest
+        strategy:
+            matrix:
+                php-versions: ['8.0']
+            fail-fast: false
+        name: PHP ${{ matrix.php-versions }} Test dev dependencies on ubuntu-latest
+        steps:
+            - name: Checkout
+              uses: actions/checkout@v2
+
+            - name: Setup PHP
+              uses: shivammathur/setup-php@v2
+              with:
+                  php-version: ${{ matrix.php-versions }}
+                  extensions: zip
+
+            - name: Get composer cache directory
+              id: composercache
+              run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+
+            - name: Cache dependencies
+              uses: actions/cache@v2
+              with:
+                  path: ${{ steps.composercache.outputs.dir }}
+                  key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.json') }}
+                  restore-keys: ${{ runner.os }}-composer-
+
+            - name: Allow dev dependencies
+              run: composer config minimum-stability dev
+
+            - name: Install dependencies
+              run: composer install --prefer-dist
+
+            - name: Run tests
+              run: vendor/bin/simple-phpunit
+
     phpunit-lowest:
         runs-on: ubuntu-latest
         name: PHP 8.0 (lowest) Test on ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,7 @@ jobs:
         runs-on: ubuntu-latest
         strategy:
             matrix:
-                php-versions: ['7.1', '7.2', '7.3', '7.4', '8.0']
+                php-versions: ['8.0']
             fail-fast: false
         name: PHP ${{ matrix.php-versions }} Test on ubuntu-latest
         steps:
--- a/src/DomCrawler/Crawler.php
+++ b/src/DomCrawler/Crawler.php
@@ -86,11 +86,7 @@ final class Crawler extends BaseCrawler implements WebDriverElement
         throw $this->createNotSupportedException(__METHOD__);
     }
 
-    /**
-     * @return static
-     */
-    #[\ReturnTypeWillChange]
-    public function eq($position): self
+    public function eq($position): static
     {
         if (isset($this->elements[$position])) {
             return $this->createSubCrawler([$this->elements[$position]]);
@@ -109,20 +105,12 @@ final class Crawler extends BaseCrawler implements WebDriverElement
         return $data;
     }
 
-    /**
-     * @return static
-     */
-    #[\ReturnTypeWillChange]
-    public function slice($offset = 0, $length = null): self
+    public function slice($offset = 0, $length = null): static
     {
         return $this->createSubCrawler(\array_slice($this->elements, $offset, $length));
     }
 
-    /**
-     * @return static
-     */
-    #[\ReturnTypeWillChange]
-    public function reduce(\Closure $closure): self
+    public function reduce(\Closure $closure): static
     {
         $elements = [];
         foreach ($this->elements as $i => $element) {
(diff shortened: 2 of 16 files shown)


jordisala1991:
This one is strange:

```
PHP Fatal error:  Declaration of Symfony\Component\Panther\DomCrawler\Crawler::eq($position): Symfony\Component\Panther\DomCrawler\Crawler must be compatible with Symfony\Component\DomCrawler\Crawler::eq(int $position): Symfony\Component\ErrorHandler\DebugClassLoader in /home/runner/work/panther/panther/src/DomCrawler/Crawler.php on line 89
```

jordisala1991:
IMO the contents of this commit: https://github.com/symfony/panther/pull/509/commits/31d379ed84cc7cb041ce2e4ebedeeda448d0a2b3 should be reverted, because it makes this bundle not compatible with php < 8.0

@wouterj do you agree? Should I open one PR to revert this return typehint? Are multiple PRs needed or just one (IMO it should be better in one, because all of them have to be accepted for this PR to continue, otherwise php < 8 have to be dropped to accept Symfony 6)?

src/WebTestAssertionsTrait.php
```diff
@@ -275,9 +274,9 @@ private static function findElement(string $locator): WebDriverElement
      * @param array $options An array of options to pass to the createKernel method
      * @param array $server  An array of server parameters
      *
-     * @return AbstractBrowser A browser instance
+     * @return KernelBrowser A browser instance
      */
-    protected static function createClient(array $options = [], array $server = []): AbstractBrowser
+    protected static function createClient(array $options = [], array $server = []): KernelBrowser
```
jordisala1991, line 279: This one I am not competely sure if it is correct on Symfony side, but without changing this I got errors too. @dunglas do you know if this change is correct?

weaverryan: as far as I can tell, `KernelBrowser` extends `HttpKernelBrowser` extends `AbstractBrowser`, and this has existed since at least 4.4. So changing this to `KernelBrowser` is just "correct: (`WebTestCase::createClient()` has always had this as its `@return`) and it's just more specific: we're narrowing the return type.

src/DomCrawler/Crawler.php
```diff
@@ -86,7 +86,7 @@ public function addNode(\DOMNode $node): void
         throw $this->createNotSupportedException(__METHOD__);
     }
 
-    public function eq($position): self
+    public function eq($position): static
```
weaverryan, line 89: This is another spot where we can't do this... unless we're planning to drop support for PHP 7. But... maybe we are??? In order to support Symfony 6?

jordisala1991: I guess the core team of Symfony should take a decision on this. I can work on both implementations. Keeping support for older php version needs main Symfony repository to be adapted before 6.0 comes out. Removing support for older php versions might need a major version of this package.

jordisala1991: But indeed this one I missed when reporting here: https://github.com/symfony/symfony/issues/43021

This will be a ton of changes, because all of. this class changed its return types. 😱

chalasr: Either we remove the return types from core, or we add some conditional classes/traits declarations in this bundle. 
I'm personally ok to revert the addition of these type declarations, we can re-add them later.

weaverryan: So, should we revert these in core? If we're going to do that, I guess we should do it as quickly as we can.

(I'm still very unfamiliar with the typing situation - so apologies for not having anything intelligent to add)

jordisala1991: Yes, I am trying to fix everything here to know what types exactly should be reverted, because I wasn't taking care of static.

My idea is:
* Prepare this PR to be able to ensure that changes on Symfony will make this PR pass
* Make a PR on Symfony core to change all types needed here
* Adapt this PR when the PR on the main repo is merged

For the moment I am having some trouble with PHPStan, since it is not a dependency of this project it got upgraded to 1.0 without notice and now it reports more errors.

weaverryan: This makes sense to me - thank you and keep us posted :)

jordisala1991: Im trying a little trick to patch symfony to remove its return types so we can ensure that those changes applied to Symfony 6.0 will make build pass with all versions here.

So far I think it will work, but I am having some last troubles with PHPStan.

The idea is to remove the last 2 commits from this PR once I do the Symfony PR to remove the return typehints

wdyt?

jordisala1991:
I don't think the failure is related to my changes.

weaverryan (approved):
I'll need someone more familiar with the "types" topic on the Symfony core to take a look, but this makes sense to me - very nice work!

nicolas-grekas:
What about bumping panther to v2?

jordisala1991:
> What about bumping panther to v2?

The work to be done is mostly the same, removing only the last commit cleans up the return types to be compatible only with php 8 and sf 6.

There are some things to consider
- What other changes are needed for v2? Maybe deprecate more things or cleaning up some stuff.
- The way forward for bundle maintainers will be requiring ^1 and ^2 if they want to be compatible with php 7 and sf 5

Maybe there are more things that I cant see because I dont know a lot of this bundle.

nicolas-grekas:
For this bundle I agree. For for Symfony that'd allow withdrawing symfony/symfony#44005, which could be nice.
/cc @dunglas WDYT?

chalasr:
Works for me. Should we maintain a 1.x branch?

dunglas:
I've no strong opinion about this. I'm ok with both options.

jordisala1991:
So, which are the next steps?

weaverryan:
I’m hearing a slight preference to make a v2 of this bundle. That would be my vote as well so we can move things forward.

As far as a 1.x branch? We could decide on that later, but there is little harm to create one.

chalasr:
@jordisala1991 Could you revert the DomCrawler patch to make the code PHP8+ only, then mark this PR as ready?

jordisala1991:
Done @chalasr

.github/workflows/ci.yml
```diff
@@ -88,6 +88,43 @@ jobs:
...
+
+            - name: Cache dependencies
+              uses: actions/cache@v2
+              with:
+                  path: ${{ steps.composercache.outputs.dir }}
+                  key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.json') }}
+                  restore-keys: ${{ runner.os }}-composer-
+
+            - name: Allow dev dependencies
+              run: composer config minimum-stability dev
```
weaverryan, line 120: Normally we have `minimum-stability: "dev"` in `composer.json` by default... and that's what we always use. If we did that, then we wouldn't need this second job, I think. For example: https://github.com/symfony/string/blob/d70c35bb20bbca71fc4ab7921e3c6bda1a82a60c/composer.json#L39

weaverryan: A little more info. What we normally do is:

* Put `"minimum-stability": "dev"` directly in `composer.json`
* For "normal" builds (e.g. released versions like Symfony 5.3), we use `--prefer-stable` with  `composer update`.
* For non-released builds (e.g. Symfony 6), we would omit the `--prefer-stable`. 

BUT, I think these are not-important details. The current builds DO test the correct stuff. I vote we merge this, and then we could potentially optimize this build file later to reduce duplication.

weaverryan (approved):
I'm good with this - for a 2.0 release.

A CHANGELOG entry for 2.0 would be great.

chalasr (approved):

.github/workflows/ci.yml
jordisala1991: I think it would be nice to have a better testing of different versions, as you said it can be done on another Pr. I can bring some of the workflows we use at Sonata if you dont have a standard way for all sf repositories.

.github/workflows/ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
             - name: Setup PHP
               uses: shivammathur/setup-php@v2
               with:
-                  php-version: '7.4' # PHP CS Fixer isn't compatible with PHP 8 yet https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4702
+                  php-version: '8.0'
```
dunglas, line 18: If we don't test with PHP 7 anymore, we should bump the minimum PHP version in `composer.json`

dunglas (approved):

.github/workflows/ci.yml
jordisala1991: Done! I missed that part.

weaverryan:
@dunglas I believe you're needed to merge and tag? I would love that - it would unblock a Symfony6 release for symfony/ux... which would unblock some other symfony/ux stuff :).

dunglas:
Merged! Thanks @jordisala1991!

@weaverryan I'll try to craft a release on Monday.

Issue #481, opened 2021-07-08
The automatic naming of screenshots creates a weird directory structure


jordygroote:
Hi,

When i take a screenshot the directory is being weird because my error message contains quotes.
![Schermafbeelding 2021-07-08 om 10 50 19](https://user-images.githubusercontent.com/8653367/124892629-4fd29f00-dfda-11eb-9d32-80e51f990528.png)

In `Symfony\Component\Panther\ServerExtension:takeScreenshots`

$test contains:
`"App\Tests\Smoke\SmokeTest::testSitemapResults with data set "/" ('/')"`

![Schermafbeelding 2021-07-08 om 10 57 27](https://user-images.githubusercontent.com/8653367/124893829-61687680-dfdb-11eb-8e32-48da37190326.png)

```
strtr($test, ['\\' => '-', ':' => '_']),
```

Doesn't handle quotes, are there more people having problems with this?

trakos:
It's not caused by quote, but simply by having slashes (`/`) in data set name.

Maybe panther could use something like `Symfony\Component\String\Slugger\AsciiSlugger` from string component instead of using `strtr`? I'd imagine there's more characters that could create issues, especially on windows filesystems.

PR #510, opened 2021-11-03
Fix screenshot name for tests with data provider

trakos:
Fixes https://github.com/symfony/panther/issues/481

When using panther for tests with data provider, I've noticed that screenshots are not being made. While debugging I've noticed that there's an exception that's ignored by phpunit:

```
file_put_contents(data/panther_screenshots/2021-11-03_05-26-28_failure_AcmetTest-EndToEnd-TestCases-Orders-Admin-AddOrderTest__testAddOrder with data set "europe" ('TESTCLIENTAT', AcmeTest-EndToEnd-Library-Modules-Orders-Admin-Add-Models-Initial-TransferObject-BillingInformation Object (...), AcmeTest-EndToEnd-Library-Modules-Orders-Admin-Add-Models-Initial-TransferObject-ShippingAddress Object (...), 'PayU - transfer', 'Courier')-0.png): failed to open stream: File name too long
```

Apparently phpunit attaches all arguments to the test name when using `toString` method. To avoid that, we would have to use some methods marked as internal that might change without deprecation period.  So, instead, I've added slugification of the test name, and cut it to the first 150 characters (with all the other segments total filename length is now max 235 characters).

### DIFF
--- a/src/ServerExtension.php
+++ b/src/ServerExtension.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Symfony\Component\Panther;
 
+use Facebook\WebDriver\WebDriver;
 use PHPUnit\Runner\AfterLastTestHook;
 use PHPUnit\Runner\AfterTestErrorHook;
 use PHPUnit\Runner\AfterTestFailureHook;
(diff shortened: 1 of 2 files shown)


src/ServerExtension.php
```diff
@@ -83,17 +84,27 @@ public static function takeScreenshots(string $type, string $test): void
...
-            );
-            $client->takeScreenshot($screenshotPath);
+            $screenshotPath = self::takeClientScreenshot($_SERVER['PANTHER_ERROR_SCREENSHOT_DIR'], $type, $test, $client, $i);
             if ($_SERVER['PANTHER_ERROR_SCREENSHOT_ATTACH'] ?? false) {
                 printf('[[ATTACHMENT|%s]]', $screenshotPath);
             }
         }
     }
+
+    public static function takeClientScreenshot(string $screenshotDirectory, string $type, string $test, WebDriver $client, int $clientNumber): string
```
trakos, line 94: I've extracted this to separate method to allow testing it.

src/ServerExtension.php
```diff
@@ -83,17 +84,27 @@ public static function takeScreenshots(string $type, string $test): void
...
+    {
+        $testSlug = strtr($test, ['\\' => '-', ':' => '_']);
+        $testSlug = substr(preg_replace("/[^-_A-Za-z0-9 ]/", '', $testSlug), 0, 150);
+        $screenshotPath = sprintf('%s/%s_%s_%s_%s-%d.png',
+                                  $screenshotDirectory,
+                                  date('Y-m-d_H-i-s'),
+                                  $type,
+                                  $testSlug,
+                                  md5($test),
```
trakos, line 103: With my new slugification, some tests that are run in the same second could end up with the same filename. For example consider:

- `AcmeTest__testFile with data set ą`
- `AcmeTest__testFile with data set ę`
- `AcmeTest__testFile with data set !`
- `AcmeTest__testFile with data set @`

That last special character would be omitted, so all those tests would have the same name, as long as they run within the same second. 

That's why I'm adding md5 of the `$test` before the slugification to the filename.

dunglas (approved):

trakos:
I've fixed coding style violation (double quotes instead of single quotes).

trakos:
I've rebased and changed target to 1.x

dunglas:
Thanks!

trakos:
@tboyer-debuT my change was only merged into 1.x branch, you can see it here: https://github.com/symfony/panther/blob/a923ff6d1f6a8034e9ae50d2c186bda1a36d951f/src/ServerExtension.php#L97

It removes slashes from the string, so they no longer break tests. I suppose it should be incorporated into main branch as well, so that it would be available in version 2 as well, but I don't know if that's planned.

Right now it hasn't been released yet, so it's available only on `dev-1.x`.

PR #511, opened 2021-11-03
Fix phpstan violations

trakos:
Just some simple fixes to fix CI errors.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -19,5 +19,4 @@ parameters:
 		- '#Call to an undefined method DOMNode::getTagName\(\)\.#'
 		- '#Return type \(void\) of method Symfony\\Component\\Panther\\DomCrawler\\Crawler::clear\(\) should be compatible with return type \(Facebook\\WebDriver\\WebDriverElement\) of method Facebook\\WebDriver\\WebDriverElement::clear\(\)#'
 		- '#Method Symfony\\Component\\Panther\\DomCrawler\\Crawler::getIterator\(\) should return ArrayIterator<int, DOMNode> but returns ArrayIterator<\(int\|string\), Facebook\\WebDriver\\WebDriverElement>\.#'
-		- '#^Expression on left side of \?\? is not nullable\.$#'
 		- '#^PHPDoc type array of property Symfony\\Component\\Panther\\DomCrawler\\Field\\FileFormField\:\:\$value is not covariant with PHPDoc type string of overridden property Symfony\\Component\\DomCrawler\\Field\\FormField\:\:\$value\.$#'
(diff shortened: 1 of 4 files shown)


src/ServerTrait.php
```diff
@@ -35,7 +35,7 @@ private function stopWebServer(): void
     private function pause($message): void
     {
         if (\in_array('--debug', $_SERVER['argv'], true)
-            && $_SERVER['PANTHER_NO_HEADLESS'] ?? false
+            && ($_SERVER['PANTHER_NO_HEADLESS'] ?? false)
```
trakos, line 38: Violations of this type were actually a bug - `true && $var ?? false` is equivalent to `(true && $var) ?? false` instead of `true && ($var ?? false)`.

phpstan.neon
```diff
@@ -17,3 +17,6 @@ parameters:
 		- '#Call to an undefined method DOMNode::getTagName\(\)\.#'
 		- '#Return type \(void\) of method Symfony\\Component\\Panther\\DomCrawler\\Crawler::clear\(\) should be compatible with return type \(Facebook\\WebDriver\\WebDriverElement\) of method Facebook\\WebDriver\\WebDriverElement::clear\(\)#'
 		- '#Method Symfony\\Component\\Panther\\DomCrawler\\Crawler::getIterator\(\) should return ArrayIterator&iterable<DOMNode> but returns ArrayIterator<\(int\|string\), Facebook\\WebDriver\\WebDriverElement>\.#'
+		- '#PHPDoc type array of property Symfony\\Component\\Panther\\DomCrawler\\Field\\FileFormField::\$value is not covariant with PHPDoc type string of overridden property Symfony\\Component\\DomCrawler\\Field\\FormField::\$value#'
```
trakos, line 20: I don't really get this one, in parent class method `getValue` is type-hinted as `array|string`, but field `$value` is type-hinted as `string` exclusively. We could make `$value` hold a json-encoded array or something, but I feel like it's some inconsistency to be fixed in parent symfony class.

src/DomCrawler/Crawler.php
```diff
@@ -38,7 +38,7 @@ public function __construct(array $elements = [], ?WebDriver $webDriver = null,
     {
         $this->uri = $uri;
         $this->webDriver = $webDriver;
-        $this->elements = $elements ?? [];
+        $this->elements = $elements;
```
trakos, line 41: Since `$elements` is type-hinted as `array`, it can't be null.

dunglas (approved):
Could you target the 1.x branch and rebase, please?

trakos:
@dunglas I've rebased on 1.x and changed target.

It seems that MR https://github.com/symfony/panther/pull/521 was merged first, and it had similar changes. Most of my changes were now obsolete. However, one error was ignored instead of being fixed, so I think that it's still worth merging that.

On master it was fixed as part of symfony upgrade and cleanup.

dunglas (approved):

dunglas:
Thanks!!

PR #522, opened 2021-11-30
Remove polyfill

derrabus:
On the main branch, we require PHP 8 or higher. I think, it's therefor safe to remove the PHP 7.2 polyfill.


chalasr:
Good catch, thanks @derrabus.

PR #523, opened 2021-12-01
Fix accessing `PantherTestCaseTrait::$webServerDir` before initialization

kbond:
`PantherTestCaseTrait::getWebServerDir()` can access `PantherTestCaseTrait::$webServerDir` before it's initialized. I believe this was just missed in #517.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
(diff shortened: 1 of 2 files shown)


chalasr (approved):

chalasr:
Good catch, thanks @kbond.

PR #524, opened 2021-12-10
Added support for `matches()` and `closest()` in Crawler

pkruithof:
The `Crawler` class from the DomCrawler component defines two methods that are not implemented here, but don't throw an exception that they are not implemented. This was not obvious in my application and it took me a while to figure out why: because the DomCrawler works with Dom nodes, but Panther's crawler can also use web elements (when using the web driver).

This PR adds support for both methods. I checked if there are any other methods that were missing implementations, but these seem to be the last ones. The test cases I've copied from the DomCrawler component and modified to work with this test suite.


tests/DomCrawler/CrawlerTest.php
```diff
@@ -156,6 +156,50 @@ public function testSiblings(callable $clientFactory): void
...
+        $this->assertTrue($p->matches('.foo'));
+        $this->assertFalse($p->matches('#other-id'));
+        $this->assertFalse($p->matches('div'));
+        $this->assertFalse($p->matches('.bar'));
+    }
+
+    /**
+     * @dataProvider clientFactoryProvider
+     */
+    public function testClosest(callable $clientFactory): void
```
pkruithof, line 178: Test and fixture copied from https://github.com/symfony/dom-crawler/blob/5.4/Tests/AbstractCrawlerTest.php#L937

pkruithof:
The failing test seems to be random and not related to this PR.

dunglas (approved):

dunglas:
Thanks @pkruithof

PR #526, opened 2021-12-21
Fix setValue on checkbox and select fields if they are multiple

robertfausk:
Currently ```ChoiceFormField::setValue``` is only able to add new values if the field is ```select``` or ```choice```.

This only-adding behavior:
- prevents the user from deselecting a value.
- prevents the user from 
- is not conform with ```ChoiceFormField::setValue``` of ```symfony-dom-crawler``` which resets the value of the field by the given value.

This PR unsets all previous values from given field before the given value is set.

PS: Do I have do take any action to get this PR into the ```main``` branch? :upside_down_face:

### DIFF
--- a/src/DomCrawler/Field/ChoiceFormField.php
+++ b/src/DomCrawler/Field/ChoiceFormField.php
@@ -135,6 +135,9 @@ final class ChoiceFormField extends BaseChoiceFormField
             return;
         }
 
+        if ($this->selector->isMultiple()) {
+            $this->selector->deselectAll();
+        }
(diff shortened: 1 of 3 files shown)


dunglas (approved):
LGTM

dunglas:
Thanks Robert!

robertfausk:
@dunglas I just realized that this PR is not upstream merged into the ```main``` branch. By now it is only available in ```1.x```
Shall I open a new PR for this or will you take it into account?

Issue #502, opened 2021-09-29
Doesn't work with bootstrap 5 because of scroll


bastien70:
Hi, i just migrated from bootstrap 4 to bootstrap 5, and something weird is happening.

When I run a test with Panther, and it wants to go to an element on the page, it gradually scrolls until it gets there, and this results in this error:

```
Facebook\WebDriver\Exception\ElementClickInterceptedException : element click intercepted: Element is not clickable at point (570, 1407)
  (Session info: chrome=93.0.4577.63)
```

In fact the rest of the test is carried out before it has completely finished scrolling, which means that the click will completely bug.

When I was using bootstrap 4 it would scroll sharply to the page element, and I didn't have this problem

mathieu-dumoutier:
I have exactly the same behavior after a migration Bootstrap 4 to Bootstrap 5 on many tests.

domis86:
can you try scrolling with javascript and then clicking?

mathieu-dumoutier:
Yes I tried and I can see button with scrolling but when I use the click() method the scroll move and trigger the error message.

adrienfr:
Hi, we have the same issue with [Behat](https://github.com/Behat/Behat) when migrating from Bootstrap 4 to Bootstrap 5

lwillems:
Got the same issue with panther and boostrap5
I fixed it temporaly by using location of my attribute and javascript scrolling

``` php
$client->executeScript("window.scrollTo({$attribute->getLocation()->getX()}, {$attribute->getLocation()->getY()})");
```

dunglas:
In my opinion this should be reported to Bootstrap, and documented in the readme. But we'll not implement non-standard behaviors of third party libraries directly in Panther.

adrienfr:
Thanks @dunglas! 
I just found this on Bootstrap's Github, https://github.com/twbs/bootstrap/issues/35536#issuecomment-992327098, checking if this fix the issue right now

adrienfr:
I can confirm this solution fix my issues with Behat on Bootstrap 5+:
```css
html {
    scroll-behavior: auto !important;
}

```

mathieu-dumoutier:
I confirm @adrienfr's solution solved many tests break down until here.

bastien70:
I take this opportunity to provide the solution I found at the time (sorry I had completely forgotten this issue).

I just modified a bootstrap 5 variable like this to disable smooth scroll, in my .scss file

```scss
$enable-smooth-scroll: false;
```

[...]

PR #531, opened 2022-01-19
Updated README.md to allow Bootstrap 5 problem

bastien70:
See #502 

I added a **Troubleshooting** section and the fix for Bootstrap 5 problem

### DIFF
--- a/README.md
+++ b/README.md
@@ -677,6 +677,18 @@ The following features are not currently supported:
 
 Pull Requests are welcome to fill the remaining gaps!
 
+## Troubleshooting
+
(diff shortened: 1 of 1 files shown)


dunglas (approved):

dunglas:
Thanks!

arderyp:
Thanks for merging this.  You might consider adding another approach:

```
// call this after crawling to a page if you'd rather not override Bootstrap's default scrolling behavior
$client->executeScript('document.documentElement.style.scrollBehavior = "auto";');
```

PR #537, opened 2022-03-16
Add ability to customize HttpClient and Panther Client

HypeMC:
Similar to #437 this will add the ability to pass options to the HttpClient when using the HttpBrowser.
This can be useful when an external web server (via `PANTHER_EXTERNAL_BASE_URI`) is used.


dunglas (approved):

nicolas-grekas (approved):

dunglas:
Thanks!

PR #545, opened 2022-04-21
Implement coverage support

mpiot:
This PR try to implement Code Coverage support by using Panther when testing. The logic come from solutions given in #8, with some adaptation to be used with many Coverage Drivers: xdebug, pcov and phpdbg.

Actually I've not implements tests, just copy/paste my project implementation for review with some adaptations.

### DIFF
--- a/src/PantherTestCase.php
+++ b/src/PantherTestCase.php
@@ -18,6 +18,7 @@ use PHPUnit\Framework\TestResult;
 use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
 use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\Finder\Finder;
+use Symfony\Component\Finder\SplFileInfo;
 
 if (class_exists(WebTestCase::class)) {
     abstract class PantherTestCase extends WebTestCase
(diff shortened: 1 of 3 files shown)


src/PantherTestCase.php
```diff
@@ -23,6 +26,39 @@ abstract class PantherTestCase extends WebTestCase
...
+            }
+
+            if (!class_exists(Finder::class)) {
+                throw new \LogicException('The Finder component is not installed. Try running "composer require --dev symfony/finder".');
+            }
+
+            /** @var \SplFileInfo[] $files */
+            $files = (new Finder())->in(self::COVERAGE_DIRECTORY)->files()->name('*.code_coverage');
+            foreach ($files as $file) {
+                $content = file_get_contents($file->getRealPath());
```
HypeMC, line 50: The Finder component's [`getIterator()` method](https://github.com/symfony/symfony/blob/337426c43ae0d22f4bafaf9cbf8849dea1310c57/src/Symfony/Component/Finder/Finder.php#L611) actually returns an instance of [`Symfony\Component\Finder\SplFileInfo`](https://github.com/symfony/symfony/blob/337426c43ae0d22f4bafaf9cbf8849dea1310c57/src/Symfony/Component/Finder/SplFileInfo.php#L19) which has a [`getContents` method](https://github.com/symfony/symfony/blob/337426c43ae0d22f4bafaf9cbf8849dea1310c57/src/Symfony/Component/Finder/SplFileInfo.php#L74), maybe it would make sense to utilize it instead:

```suggestion
            $files = (new Finder())->in(self::COVERAGE_DIRECTORY)->files()->name('*.code_coverage');
            foreach ($files as $file) {
                $content = $file->getContents();
```

mpiot: You’re right, I’ll change it 😀

src/PantherTestCase.php
```diff
@@ -23,6 +27,39 @@ abstract class PantherTestCase extends WebTestCase
 
         public const CHROME = 'chrome';
         public const FIREFOX = 'firefox';
+        public const COVERAGE_DIRECTORY = __DIR__.'/../var/panther-coverage';
+
+        public function run(TestResult $result = null): TestResult
```
Chris53897, line 32: This will not work anymore with PHPUnit 10 ;(
Now it is `final public function run(): void`

If someone found a solution to get it work with PHPUnit 10, please let us know.

PR #550, opened 2022-06-06
Allow input type checkbox without need to set attribute "name"

robertfausk:
Currently all checkboxes must have a ```name``` attribute to be usable with ```symfony/panther``` and the ```ChoiceFormField```.

This PR allows instances of WebDriverCheckboxes without a ```name``` attribute, e.g.:
```
<input type="checkbox" value="noNameCheckbox">
```


dunglas (approved):

dunglas:
Thank you!

PR #556, opened 2022-06-16
Explicit string variable

MathiasReker:
Converts implicit variables into explicit ones in double-quoted strings or heredoc syntax.

### DIFF
--- a/src/ProcessManager/WebServerReadinessProbeTrait.php
+++ b/src/ProcessManager/WebServerReadinessProbeTrait.php
@@ -55,7 +55,7 @@ trait WebServerReadinessProbeTrait
 
             if (Process::STATUS_STARTED !== $status) {
                 if (microtime(true) - $start >= $timeout) {
-                    throw new \RuntimeException("Could not start $service (or it crashed) after $timeout seconds.");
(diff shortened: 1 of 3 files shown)


dunglas (approved):

src/ProcessManager/WebServerReadinessProbeTrait.php
```diff
@@ -55,7 +55,7 @@ public function waitUntilReady(Process $process, string $url, string $service, b
 
             if (Process::STATUS_STARTED !== $status) {
                 if (microtime(true) - $start >= $timeout) {
-                    throw new \RuntimeException("Could not start $service (or it crashed) after $timeout seconds.");
+                    throw new \RuntimeException("Could not start ${service} (or it crashed) after ${timeout} seconds.");
```
renanbr, line 58: it could be better to use `{$service}`, `{$timeout}`, and `{$message}` everywhere for [future compatibility](https://wiki.php.net/rfc/deprecate_dollar_brace_string_interpolation)

dunglas:
Shouldn't we handle all these CS issues directly in PHP CS Fixer, as we do for other SF components? cc @nicolas-grekas

PR #565, opened 2022-08-21
fix: ignore curl exceptions when closing webdriver inside destructor

mtorromeo:
Curl exceptions when quitting WebDriver are being thrown for different reasons, but are hard to handle since the method is called from inside the destructor.

The destructor is called when we are done with WebDriver anyway, and if there is an error quitting (most likely because WebDriver is not responding) we can safely ignore it IMO.

The try/catch is placed on purpose on the destructor only and not in the quit method so that when calling quit manually the exception can still be handled by the caller.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -15,6 +15,7 @@ namespace Symfony\Component\Panther;
 
 use Facebook\WebDriver\Exception\NoSuchElementException;
 use Facebook\WebDriver\Exception\TimeoutException;
+use Facebook\WebDriver\Exception\WebDriverCurlException;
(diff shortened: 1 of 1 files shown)


Chris53897:
@mtorromeo Thank you for working on this.

Looks like this works under windows, but not on mac (according to unittests).

I am using https://github.com/zenstruck/browser (on Mac) and with your changes i get this message.
I guess the same error as in the failing unittests for mac.
```
Call to a member function append() on null
Process finished with exit code 2
```

dunglas:
Thank you!

Issue #283, opened 2020-01-16
Panther with Authentication


Sethbass:
Hello,

First I would like to thank you to improve this part of symfony and helping us to get rid of Behat.

I have created a symfony app which needs authentication. This means that / is protected by the firewall main and every URL needs an authenticated user. 
On top of that I have custom authenticator which extends AbstractGuardAuthenticator. The authentication process is based on CAS (not my choice I have to cope with it).

I have tried to start my functional tests using Panther and followed the documentation here :
https://github.com/symfony/panther

So far so good :) Then I have to deal immediately with the authentication, so I found this tutorial : 
https://symfony.com/doc/4.3/testing/http_authentication.html

 First thing I have discovered is that I need to visit a page on my website otherwise I get this error :

``` 
Facebook\WebDriver\Exception\InvalidCookieDomainException: invalid cookie domain
  (Session info: chrome=79.0.3945.117)
  (Driver info: chromedriver=78.0.3904.70 (edb9c9f3de0247fd912a77b7f6cae7447f6d3ad5-refs/branch-heads/3904@{#800}),platform=Mac OS X 10.15.2 x86_64)
```

Ok then I have created a test page accessible anonymously and got rid of the error. Not sure this is a best practice.

Then I was able to run the following code : 
```
<?php

namespace App\Tests;

use App\Entity\User;
use Symfony\Component\BrowserKit\Cookie;
use Symfony\Component\Panther\PantherTestCase;
use Symfony\Component\Security\Guard\Token\PostAuthenticationGuardToken;

class HomeControllerTest extends PantherTestCase
{
    private $client = null;

    public function setUp()
    {
        $this->client = static::createPantherClient();
    }

    public function testHome()
    {

        $this->logIn();
        $crawler = $this->client->request('GET', '/test');

        sleep(5);
        $user = self::$container->get('security.helper')->getUser();
        echo ('The user is :'.$user);
        $crawler = $this->client->request('GET', '/');

    }

    private function logIn()
    {

        $doctrine = self::$container->get('doctrine.orm.default_entity_manager');

        /** @var User $user */
        $user = $doctrine->getRepository(User::class)->findOneBy(['email' => 'sebastien@code-it.fr']);

        $session = self::$container->get('session');


        // you may need to use a different token class depending on your application.
        // for example, when using Guard authentication you must instantiate PostAuthenticationGuardToken
        $token = new PostAuthenticationGuardToken($user, 'main', $user->getRoles());
[...]

PR #574, opened 2022-10-07
Add `loginUser()` helper method to Panther client

plotbox-io:
Please see [here](https://github.com/symfony/panther/issues/283) for discussion context

When doing E2E testing, it is common to need to login a user to perform certain actions. We often don't care about the login process for 99% of our tests, but we must still perform the steps each time. This can add unneccesary time and complexity onto the tests that could be avoided if only we had a way to simulate logins.

It would be nice to have a helper function named loginUser() similar to the [Application Test](https://symfony.com/doc/current/testing.html#application-tests) client. However, as we are doing E2E testing, we know that there is an added complexity of the environment settings of the alternate webserver (specifically how it handles session data). So, to solve this, we can pass SessionStorageInterface and firewallContext as arguments (but with sensisble defaults so they could be omitted in most cases).

Example of calling code would be:
```
$panther->logInUser(
    $user,
    new MockFileSessionStorage('/app/var/cache/panther/sessions'),
    'main'
);
```

Or simple (for most cases)

```
$panther->logInUser($user);
```

Have tested this approach (in userland code) and is working well for me. Was suggested in the discussion that I raise a PR for integration to Panther library code so here it is for your consideration. Please let me know if you need anything more from me.

Richard.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -541,7 +541,7 @@ final class Client extends AbstractBrowser implements WebDriver, JavaScriptExecu
         // set a cookie to the uninitialised web driver. The path does not matter for this
         $this->request('GET', '/');
 
-        if(null === $sessionStorage) {
+        if (null === $sessionStorage) {
             $sessionStorage = new MockFileSessionStorage('/app/var/cache/panther/sessions');
         }
         $session = new Session($sessionStorage);
--- a/src/Client.php
+++ b/src/Client.php
@@ -29,12 +29,16 @@ use Facebook\WebDriver\WebDriverOptions;
 use Facebook\WebDriver\WebDriverTargetLocator;
 use Facebook\WebDriver\WebDriverWait;
 use Symfony\Component\BrowserKit\AbstractBrowser;
+use Symfony\Component\BrowserKit\Cookie;
 use Symfony\Component\BrowserKit\History;
 use Symfony\Component\BrowserKit\Request;
 use Symfony\Component\BrowserKit\Response;
 use Symfony\Component\DomCrawler\Crawler;
 use Symfony\Component\DomCrawler\Form;
 use Symfony\Component\DomCrawler\Link;
+use Symfony\Component\HttpFoundation\Session\Session;
+use Symfony\Component\HttpFoundation\Session\Storage\MockFileSessionStorage;
+use Symfony\Component\HttpFoundation\Session\Storage\SessionStorageInterface;
 use Symfony\Component\Panther\Cookie\CookieJar;
 use Symfony\Component\Panther\DomCrawler\Crawler as PantherCrawler;
 use Symfony\Component\Panther\DomCrawler\Form as PantherForm;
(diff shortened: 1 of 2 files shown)


shadydealer:
@plotbox-io
Some apps would require a token other than a `UsernamePasswordToken`, could you make it an optional parameter instead?

Also wondering if this PR is even being considered?

Bilge:
The sheer number of assumptions baked into this is truly staggering.

src/Client.php
```diff
@@ -513,6 +519,42 @@ public function getWebDriver(): WebDriver
...
+            throw new \LogicException(sprintf('"%s" requires symfony/security-core to be installed.', __METHOD__));
+        }
+
+        if (!$user instanceof UserInterface) {
+            throw new \LogicException(sprintf('The first argument of "%s" must be instance of "%s", "%s" provided.', __METHOD__, UserInterface::class, get_debug_type($user)));
+        }
+
+        // Make a single request to avoid 'Invalid cookie domain' error when attempting to
+        // set a cookie to the uninitialised web driver. The path does not matter for this
+        $this->request('GET', '/');
```
dunglas, line 542: It would be nice to be able to prevent this, at least if WebDriver is already initialized.

src/Client.php
```diff
@@ -513,6 +519,42 @@ public function getWebDriver(): WebDriver
...
+
+        if (null === $sessionStorage) {
+            $sessionStorage = new MockFileSessionStorage('/app/var/cache/panther/sessions');
```
dunglas, line 545: This should something like `tempnam(sys_get_temp_dir(), 'panther-sessions');` instead. This hardcoded directory may not be writable.

dunglas (changes requested):
I'm not fond of the general approach. This will break in many cases (Panther on the host, the Symfony app in a Docker container). Also, this creates a hard coupling between Panther (the library) and the full-stack Symfony framework.

Couldn't we try an alternative approach like trying to log the user using HTTP Basic or some other kind of standard approach instead?

arderyp:
I'm interested in other approaches if people have clever ideas. Auth is an extremely common use case, so some sort of panther support for it would be cool

dunglas:
HTTP Basic?

shadydealer:
We use form login, accross all of our apps, so I don't think HTTP Basic would be of much use to us, unfortunately. 

I'm open to ideas on how to reduce the number of assumptions and how to decouple it from  the Symfony framework. If you look at the discussion in https://github.com/symfony/panther/issues/283 you'll see that the code was directly taken from the symfony testing framework, with a little modification, because, the perfect situation for me would be to have to only change `WebTestCase` to `PantherTestCase` and then use `static::createPantherClient()` where needed, making the transition from a WebTestCase to a PantherTestCase as seemless as possible.

Maybe the `loginUser` method could detect whether the project is using the Symfony framework or not, and behave differently depending on the result? Like, HTTP Basic for projects which aren't using Symfony and something along the lines of the code in the PR?

dunglas:
You could enable HTTP Basic only for the `panther` env. Or (or in addition), we could provide a helper to submit the standard Symfony login form using HTTP?

arderyp:
I'm also not keen on basic auth.  My actual site doesn't use it, and I want my tests to behave as closely to my live site as possible, including running through my real auth mechanism.

I think _allowing_ basic auth is fine, letting the implementer decide.  But forcing it doesn't seem appealing.

dunglas:
@arderyp that's the plan. You can already plug your custom auth logic if needed.

antoniovj1:
@plotbox-io Do you need some help with this?

arderyp:
Doesn't this issue remain, despite the original reporter not caring anymore? Should we reopen?

antoniovj1:
@arderyp I think that @plotbox-io has closed this because it's his pull request, but yes, the issue remains. Should we use this MR or should we open an independent issue?

arderyp:
I suppose we could open a new one

PR #580, opened 2022-11-28
Removing code smell with `$$args`

ThomasLandauer:


dunglas:
The error message is less descriptive than the previous one. I don't think this change is worth it.

ThomasLandauer:
The idea was to get rid of `$$arg` - which is *really* ugly.
And telling the users all three "problem areas" at once, can also be seen as advantage over telling them one by one.

Issue #588, opened 2023-02-22
Make Panther PHPUnit 10 compatible


arderyp:
Now that PHPUnit 10 is out, I'm adding this issue to the queue

arderyp:
is no one is working on this, I may be able to dedicate some time to looking into it.

@dunglas, since PHPUnit 10 is a substantial refactor, I am not yet sure how feasible it would be for Panther to support, for example, both 9 and 10.  It looks like Panther isn't actually dependent directly on `phpunit/phpunit`, but rather `symfony/phpunit-bridge`, which is also not PHPUnit 10 compatible.

Do you think work on this issue should wait until after `phpunit-bridge` supports 10?

Do you intend to attempt to support multiple PHPUnit versions, or would this issue warrant a new major `3` release and dropped support for PHPUnit <10?

arderyp:
it looks like `doctrine-test-bundle` does handle support for multiple php version, so maybe this is doable here as well: https://github.com/dmaicher/doctrine-test-bundle/blob/master/src/DAMA/DoctrineTestBundle/PHPUnit/PHPUnitExtension.php

Chris53897:
@arderyp Thanks for your hard work here. Maybe you can close this issue as completed?

Issue #585, opened 2023-01-18
Do you plan to update panther product till Symfony 6.2?


Mepcuk:
Do you plan to update panther product till Symfony 6.2?

dunglas:
Yes. We don't plan to deprecate Panther.

wouterj:
I'm not sure what you mean. As far as I can see, this package supports Symfony 6.2.
If you find anything to not work, please share details about it (ideally a reproducer or a bugfix).

arderyp:
I suspect OP may have simply been concerned by the fact that there hasn't been a release in over a year and there hasn't even been a commit to `main` in 4 months.  At least, that's the thought I had when I saw those numbers while contemplating [phpunit 10 compatibility](https://github.com/symfony/panther/issues/586)

dunglas:
The library is pretty stable is doesn't require much work. However we indeed have work for PHPUnit 10 compatibility.

arderyp:
That's fair enough.  I hope I didn't come off as judgemental, that's wasn't my intent.  I think it actually extremely commendable when packages become stable enough to not need frequent updates (and aren't endlessly chasing features that lead to a complexity spiral).  My reaction was just the knee jerk default these days when nearly every popular package is pushing updates with incredible frequency)

You all are doing great work, and it's an awesome package :)

arderyp:
> The library is pretty stable is doesn't require much work. However we indeed have work for PHPUnit 10 compatibility.

@dunglas @wouterj while the package is stable, it does still need some regular maintenance. Can we please merge to PHPUnit 10 compatibility PR to keep it stable and up to date?

https://github.com/symfony/panther/pull/589#issuecomment-1502169549

PR #589, opened 2023-03-01
Support PHPUnit 10

arderyp:
issue: #588 

**This PR _may_ not be ready yet.**  I am submitting it for review/guidance on the following:

## Maintain dependence on PHPUnit Bridge?
This PR does indeed allow Panther to work under PHPUnit 10. It should preserve backwards compatibility and allow those using PHPUnit 8 and 9 to also continue using Panther without issue.  But, in order to support 10, I had to (temporarily?) drop `symfony/phpunit-bridge` in favor of `phpunit/phpunit`, since the bridge is not yet PHPUnit 10 compatible. Obviously, we could hold off on merging this PR until the bridge is ready, at which point I would drop the direct PHPUnit dependency and put the bridge dependency back in place pre-merge.  But this raises the question, do we actually want/need to depend on `symfony/phpunit-bridge` here, or can we let the consumer decide, and perhaps they only want/use `phpunit/phpunit`? 

I could very well be missing something, but I'm not currently seeing any specific reason to depend on the bridge over PHPUnit directly--we aren't using any of it's code, its blocking us here and will continue to do so in the future, and depending on PHPUnit directly will leave it up to the consumer to decide for themself whether they want PHPUnit directly or the bridge or `symfony/test-pack` or something else.  

## Maintain support for PHPUnit <10?
We could make this a new major release (`v3.0.0`) and simply require 10+. There are no actual changes to the package between the latest `2.x` and this, so people on lower versions of PHPUnit could be directed to use `v2.x` and phpunit 10 users could be directed to use `v3.x`. Furthermore @dunglas has [noted](https://github.com/symfony/panther/issues/585#issuecomment-1435966921) that the package is stable and relatively unchanging at this point, so if we imagine continuing that trend, then it could be fine to drop <10 support in a new `v3.x` release.

If you all are okay with dropping <10 support and cutting a new release, then I should clean some stuff up here. But if you want to continue supporting multiple versions, what we have here should suffice.  It's messier and less simple/straight forward than it would be without supporting multiple versions, but I guess that's par for the course with third party packages.

## Restructuring
What I have here is the simplest "make it work" approach following Panther's current approach, and borrowing influence from [doctrine-test-bundle](https://github.com/dmaicher/doctrine-test-bundle/blob/master/src/DAMA/DoctrineTestBundle/PHPUnit/PHPUnitExtension.php)'s approach to solving this problem. I didn't want to get crazy and start restructuring things only for you all to not accept that.  The current structure I'd like to clean up is as follows:

```
src/
    ServerTrait.php
    ServerExtensionLegacy.php
    ServerExtension.php
```

I'd like to change it to:

```
src/
    PHPUnit/
        ServerExtensionTrait.php
        ServerExtension.php
        V9/
            ServerExtension.php
        V10/
            ServerExtension.php
            Subscriber/
                TestRunnerStartedSubscriber
                TestRunnerFinishedSubscriber
                TestStartedSubscriber
                TestFinishedSubscriber
                ErroredSubscriber
                FailedSubscriber
```

If we drop PHPUnit <10 support, then it would be a cleaner and simpler:

```
src/
    PHPUnit/
        ServerExtension.php
        Subscriber/
            TestRunnerStartedSubscriber
            TestRunnerFinishedSubscriber
            TestStartedSubscriber
            TestFinishedSubscriber
            ErroredSubscriber
            FailedSubscriber
```

These would all be very small and clean classes.

Both restructures would be BC breaks, as it would require updating the namespace of the extension referenced in `phpunit.xml`. We could avoid BC break by creating a class alias, but I don't think it's worth doing that. It's extra clutter that's not very useful and fixing thew BC break would be very easy for everyone.

I look forward to hearing what you think.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
+/.idea/
 /.php-cs-fixer.php
 /.php-cs-fixer.cache
+/.phpunit.cache
 /.phpunit.result.cache
 /composer.phar
 /composer.lock
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-/.idea/
 /.php-cs-fixer.php
 /.php-cs-fixer.cache
 /.phpunit.cache
--- a/composer.json
+++ b/composer.json
@@ -44,8 +44,6 @@
     "sort-packages": true
   },
   "require-dev": {
-    "friendsofphp/php-cs-fixer": "^3.14",
-    "phpstan/phpstan": "^1.10",
     "phpunit/phpunit": "^10.0",
     "symfony/css-selector": "^5.3 || ^6.0",
     "symfony/framework-bundle": "^5.3 || ^6.0",
--- a/composer.json
+++ b/composer.json
@@ -45,6 +45,7 @@
     "sort-packages": true
   },
   "require-dev": {
+    "friendsofphp/php-cs-fixer": "^3.14",
     "phpunit/phpunit": "^10.0",
     "symfony/css-selector": "^5.3 || ^6.0",
     "symfony/framework-bundle": "^5.3 || ^6.0",
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
     }
   ],
   "require": {
-    "php": ">=8.0",
+    "php": ">=8.1",
     "ext-dom": "*",
     "ext-libxml": "*",
     "php-webdriver/webdriver": "^1.8.2",
--- a/composer.json
+++ b/composer.json
@@ -44,9 +44,9 @@
     "sort-packages": true
   },
   "require-dev": {
-    "phpunit/phpunit": "^10.0",
     "symfony/css-selector": "^5.3 || ^6.0",
     "symfony/framework-bundle": "^5.3 || ^6.0",
-    "symfony/mime": "^5.3 || ^6.0"
+    "symfony/mime": "^5.3 || ^6.0",
+    "symfony/phpunit-bridge": "^5.3 || ^6.0"
   }
 }
--- a/composer.json
+++ b/composer.json
@@ -21,6 +21,7 @@
     "ext-dom": "*",
     "ext-libxml": "*",
     "php-webdriver/webdriver": "^1.8.2",
+    "phpstan/phpstan": "^1.10",
     "symfony/browser-kit": "^5.3 || ^6.0",
     "symfony/dependency-injection": "^5.3 || ^6.0",
     "symfony/deprecation-contracts": "^2.4 || ^3",
@@ -44,9 +45,9 @@
     "sort-packages": true
   },
   "require-dev": {
+    "phpunit/phpunit": "^10.0",
     "symfony/css-selector": "^5.3 || ^6.0",
     "symfony/framework-bundle": "^5.3 || ^6.0",
-    "symfony/mime": "^5.3 || ^6.0",
-    "symfony/phpunit-bridge": "^5.3 || ^6.0"
+    "symfony/mime": "^5.3 || ^6.0"
   }
 }
--- a/composer.json
+++ b/composer.json
@@ -17,11 +17,10 @@
     }
   ],
   "require": {
-    "php": ">=8.1",
+    "php": ">=8.0",
     "ext-dom": "*",
     "ext-libxml": "*",
     "php-webdriver/webdriver": "^1.8.2",
-    "phpstan/phpstan": "^1.10",
     "symfony/browser-kit": "^5.3 || ^6.0",
     "symfony/dependency-injection": "^5.3 || ^6.0",
     "symfony/deprecation-contracts": "^2.4 || ^3",
@@ -46,6 +45,7 @@
   },
   "require-dev": {
     "friendsofphp/php-cs-fixer": "^3.14",
+    "phpstan/phpstan": "^1.10",
     "phpunit/phpunit": "^10.0",
     "symfony/css-selector": "^5.3 || ^6.0",
     "symfony/framework-bundle": "^5.3 || ^6.0",
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -3,12 +3,13 @@ parameters:
 	paths:
 		- src
 		- tests
-	# @todo use below if/when we are back on symfony/phpunit-bridge
-	#bootstrapFiles:
-	#	- vendor/bin/.phpunit/phpunit/vendor/autoload.php
+	bootstrapFiles:
+		- vendor/bin/.phpunit/phpunit/vendor/autoload.php
 	inferPrivatePropertyTypeFromConstructor: true
 	excludePaths:
 	    - tests/DummyKernel.php
+	    # There are lots of missing phpunit classes since we are supporting multiple versions
+	    - src/ServerExtension.php
 	ignoreErrors:
 		# False positive
 		- '#Call to an undefined method ReflectionType::getName\(\)\.#'
@@ -17,3 +18,4 @@ parameters:
 		# Require a redesign of the underlying Symfony components
 		- '#Call to an undefined method DOMNode::getTagName\(\)\.#'
 		- '#Return type \(void\) of method Symfony\\Component\\Panther\\DomCrawler\\Crawler::clear\(\) should be compatible with return type \(Facebook\\WebDriver\\WebDriverElement\) of method Facebook\\WebDriver\\WebDriverElement::clear\(\)#'
+
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -3,8 +3,9 @@ parameters:
 	paths:
 		- src
 		- tests
-	bootstrapFiles:
-		- vendor/bin/.phpunit/phpunit/vendor/autoload.php
+	# @todo use below if/when we are back on symfony/phpunit-bridge
+	#bootstrapFiles:
+	#	- vendor/bin/.phpunit/phpunit/vendor/autoload.php
 	inferPrivatePropertyTypeFromConstructor: true
 	excludePaths:
 	    - tests/DummyKernel.php
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -18,4 +18,3 @@ parameters:
 		# Require a redesign of the underlying Symfony components
 		- '#Call to an undefined method DOMNode::getTagName\(\)\.#'
 		- '#Return type \(void\) of method Symfony\\Component\\Panther\\DomCrawler\\Crawler::clear\(\) should be compatible with return type \(Facebook\\WebDriver\\WebDriverElement\) of method Facebook\\WebDriver\\WebDriverElement::clear\(\)#'
-
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -2,30 +2,34 @@
 
 <!-- http://phpunit.de/manual/4.1/en/appendixes.configuration.html -->
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/6.3/phpunit.xsd"
+         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/10.0/phpunit.xsd"
          backupGlobals="false"
          bootstrap="vendor/autoload.php"
          colors="true"
+         cacheDirectory=".phpunit.cache"
 >
 
-    <php>
-        <env name="KERNEL_CLASS" value="Symfony\Component\Panther\Tests\DummyKernel" />
-        <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[direct]=0" />
-    </php>
+  <coverage>
+    <include>
+      <directory>.</directory>
+    </include>
+    <exclude>
+      <directory>tests</directory>
+      <directory>vendor</directory>
+    </exclude>
+  </coverage>
 
-    <testsuites>
-        <testsuite name="Project Test Suite">
-            <directory>tests</directory>
-        </testsuite>
-    </testsuites>
+  <php>
+    <env name="KERNEL_CLASS" value="Symfony\Component\Panther\Tests\DummyKernel"/>
+    <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[direct]=0"/>
+    <!--@todo uncomment below if/when back on symfony/phpunit-bridge -->
+    <!--<server name="SYMFONY_PHPUNIT_VERSION" value="10.0"/>-->
+  </php>
 
-    <filter>
-        <whitelist processUncoveredFilesFromWhitelist="true">
-            <directory>.</directory>
-            <exclude>
-                <directory>tests</directory>
-                <directory>vendor</directory>
-            </exclude>
-        </whitelist>
-    </filter>
+  <testsuites>
+    <testsuite name="Project Test Suite">
+      <directory>tests</directory>
+    </testsuite>
+  </testsuites>
+
 </phpunit>
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -2,34 +2,30 @@
 
 <!-- http://phpunit.de/manual/4.1/en/appendixes.configuration.html -->
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/10.0/phpunit.xsd"
+         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/6.3/phpunit.xsd"
          backupGlobals="false"
          bootstrap="vendor/autoload.php"
          colors="true"
-         cacheDirectory=".phpunit.cache"
 >
 
-  <coverage>
-    <include>
-      <directory>.</directory>
-    </include>
-    <exclude>
-      <directory>tests</directory>
-      <directory>vendor</directory>
-    </exclude>
-  </coverage>
+    <php>
+        <env name="KERNEL_CLASS" value="Symfony\Component\Panther\Tests\DummyKernel" />
+        <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[direct]=0" />
+    </php>
 
-  <php>
-    <env name="KERNEL_CLASS" value="Symfony\Component\Panther\Tests\DummyKernel"/>
-    <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[direct]=0"/>
-    <!--@todo uncomment below if/when back on symfony/phpunit-bridge -->
-    <!--<server name="SYMFONY_PHPUNIT_VERSION" value="10.0"/>-->
-  </php>
+    <testsuites>
+        <testsuite name="Project Test Suite">
+            <directory>tests</directory>
+        </testsuite>
+    </testsuites>
 
-  <testsuites>
-    <testsuite name="Project Test Suite">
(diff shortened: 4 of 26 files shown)


arderyp:
I am not sure how to address the failures above, if I could have guidance on that as well.

I made sure PHPStan was clean (is there a reason it's not included as a dev package?).  Also, if there a reason your coding stanrad tool is not a dev package? What should I install to lint locally?

It looks like there is an issue with supporting php 8.0... maybe the new `v3.x` release should require 8.1, which would avoid those failures.

Again, guidance would be much appreciated.

composer.json
```diff
@@ -17,10 +17,11 @@
     }
   ],
   "require": {
-    "php": ">=8.0",
+    "php": ">=8.1",
     "ext-dom": "*",
     "ext-libxml": "*",
     "php-webdriver/webdriver": "^1.8.2",
+    "phpstan/phpstan": "^1.10",
```
dunglas, line 24: this should be removed (PHPStan is not a hard dependency of this project).

composer.json
```diff
@@ -44,9 +45,9 @@
     "sort-packages": true
   },
   "require-dev": {
+    "phpunit/phpunit": "^10.0",
```
dunglas, line 48: I suggest keeping `symfony/phpunit-bridge` for now, but creating a temporary GitHub Actions workflow that will execute `composer remove symfony/phpunit-bridge` and then `composer require  phpunit/phpunit`.

So we can test Panther using the bridge (which is necessary to detect deprecations), as well as with PHPUnit 10.

We'll remove this temporary workflow when PHPUnit Bridge will support PHPUnit 10.

phpstan.neon
```diff
@@ -3,8 +3,9 @@ parameters:
 	paths:
 		- src
 		- tests
-	bootstrapFiles:
-		- vendor/bin/.phpunit/phpunit/vendor/autoload.php
+	# @todo use below if/when we are back on symfony/phpunit-bridge
```
dunglas, line 6: With the strategy previously mentioned, we can revert the changes in this file.

phpunit.xml.dist
```diff
@@ -2,30 +2,34 @@
 
 <!-- http://phpunit.de/manual/4.1/en/appendixes.configuration.html -->
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
dunglas, line 4: Are the changes in this file mandatory? If we can keep it "as-is", we ensure that the project still work as expected with older versions.

composer.json
```diff
@@ -17,10 +17,11 @@
     }
   ],
   "require": {
-    "php": ">=8.0",
+    "php": ">=8.1",
```
dunglas, line 20: This is probably not necessary if we do not require PHPUnit 10 as I suggest after.

src/PantherTestCaseTrait.php
```diff
@@ -129,14 +128,19 @@ public static function isWebServerStarted(): bool
 
     public function takeScreenshotIfTestFailed(): void
     {
-        if (!\in_array($this->getStatus(), [BaseTestRunner::STATUS_ERROR, BaseTestRunner::STATUS_FAILURE], true)) {
+        // provided by PHPUnit\Framework\TestCase
+        if (!method_exists($this, 'status')) {
```
dunglas, line 132: Could you adapt this to support both PHPUnit 10 and older versions?

src/PantherTestCaseTrait.php
```diff
@@ -239,4 +243,9 @@ private static function getWebServerDir(array $options): string
 
         return $_SERVER['PANTHER_WEB_SERVER_DIR'];
     }
+
+    private static function isGetClientStaticMethodAvailable(): bool
```
dunglas, line 247: This change doesn't look necessary. Can you please keep the diff as small as possible to help the review?

dunglas (changes requested):

dunglas:
Thanks for working on this!

* IMHO we should support both PHPUnit 10 and older versions at the same time, at least as long as Symfony supports older PHPUnit versions
* The bridge is only a dev dependency, we use it to track deprecations
* CSs are handled by "fabbot", which is basically PHP CS Fixer. Just run the command provided by the bot to fix the CS automatically.

arderyp:
I ran the fixer.  I also added `php-cs-fixer` as a dependency before seeing your comment on the phpstan dependency. I will remove both dependencies, work through your other comments here, and then check back in shortly.

composer.json
arderyp: moved to dev dependencies: https://github.com/arderyp/panther/commit/5cf344a56fbadfced1d5f538b20a805803117b3f

composer.json
arderyp: @dunglas I can do this, but the only problem is that it prevents me from being able to run tests locally.  I can remove `phpunit/phpunit` and require dev `symfony/phpunit-bridge` and use `<server name="SYMFONY_PHPUNIT_VERSION" value="10.0"/>` in `phpunit.xml.dist` but I cannot run `vendor/bin/simple-phpunit` because the bridge doesn't support phpunit 10 yet:

```
PHP Fatal error:  Uncaught Error: Class "PHPUnit\TextUI\Command" not found in /home/arderyp/panther/vendor/symfony/phpunit-bridge/Legacy/CommandForV9.php:25
Stack trace:
#0 /home/arderyp/panther/vendor/bin/.phpunit/phpunit-10.0-0/vendor/composer/ClassLoader.php(571): include()
#1 /home/arderyp/panther/vendor/bin/.phpunit/phpunit-10.0-0/vendor/composer/ClassLoader.php(428): Composer\Autoload\includeFile()
#2 [internal function]: Composer\Autoload\ClassLoader->loadClass()
#3 /home/arderyp/panther/vendor/symfony/phpunit-bridge/TextUI/Command.php(17): class_alias()
#4 /home/arderyp/panther/vendor/bin/.phpunit/phpunit-10.0-0/vendor/composer/ClassLoader.php(571): include('...')
#5 /home/arderyp/panther/vendor/bin/.phpunit/phpunit-10.0-0/vendor/composer/ClassLoader.php(428): Composer\Autoload\includeFile()
#6 /home/arderyp/panther/vendor/bin/.phpunit/phpunit-10.0-0/phpunit(22): Composer\Autoload\ClassLoader->loadClass()
#7 /home/arderyp/panther/vendor/symfony/phpunit-bridge/bin/simple-phpunit.php(441): include('...')
#8 /home/arderyp/panther/vendor/symfony/phpunit-bridge/bin/simple-phpunit(13): require('...')
#9 /home/arderyp/panther/vendor/bin/simple-phpunit(120): include('...')
#10 {main}
  thrown in /home/arderyp/panther/vendor/symfony/phpunit-bridge/Legacy/CommandForV9.php on line 25
```
see: https://github.com/symfony/symfony/issues/49069

furthermore, I haven't tested yet, but this would probably prevent me from actually being able to use Panther with PHPUnit 10.  Does what I'm saying make sense, or do you think I am misunderstanding something?

phpunit.xml.dist
arderyp: they are necessary.  I have converted our own testsuite to be PHPUnit 10 compatible (there are BC breaks).

I don't think there is any problem in our project requiring a dev dependency explicitly of phpunit 10, as our tests have been updated and now require 10.  since it is a dev dependency and not a hard dependency, consumers can pull in Panther and use whatever PHPUnit they provide in their project's requirements.

composer.json
arderyp: done: https://github.com/arderyp/panther/commit/5cf344a56fbadfced1d5f538b20a805803117b3f

src/PantherTestCaseTrait.php
arderyp: I can revert this, but there will still be associated changes, as I replaced `\is_callable()` with `method_exists()` in order to resolve 2 phpstan errors that were/are present on `symfony/panther:main`

But if you want me to remove the function and deal instead with the two line changes instead, or accept the phpstan errors in which case I can fully refert to `\is_callable()`.  Let me know @dunglas

arderyp:
okay @dunglas, I've addressed what I can and need more feedback where noted.  Please let me know.

Also, is there a reason we are excluding `composer.lock` in `.gitignore`? It makes development somewhat unstable...

src/PantherTestCaseTrait.php
arderyp: done: https://github.com/arderyp/panther/commit/e7b35205822b079227382791109779772cc8743b

@dunglas resolve if this is sufficient

dunglas:
It's a best practice for libraries.

.gitignore
```diff
@@ -1,5 +1,7 @@
+/.idea/
```
dunglas, line 1: This should be in your global gitignore file, but not here.

.gitignore
```diff
@@ -1,5 +1,7 @@
...
 /.php-cs-fixer.php
 /.php-cs-fixer.cache
+/.phpunit.cache
```
dunglas, line 4: As this changed in 10? Why not use the already ignored code `.phpunit.result.cache`?

composer.json
dunglas: Yo test locally with PHPUnit 10, you'll have to do the same steps as in the CI (then revert the changes in `composer.json` before committing).

This is definitely not ideal, but quite common an only temporary. We'll be able to remove this hack when PHPUnit Bridge will support PHPUnit 10.

As PHPUnit is not a hard dependency of Panther, this will not prevent users to use Panther with PHPUnit 10 in their own projects.

TL;DR: restore the content of Panther's `composer.json` and change it only in a specific CI workflow.

phpunit.xml.dist
dunglas: It's an optional dependency, but not only a dev one. We must ensure in our CI that Panther still run on projects using PHPUnit <10.

arderyp: I'm not sure how you want to proceed here.  In order to make Panther's internal tests phpunit 10 compliant, I had to make various changes you'll see to the test files.  additionally, I had to migrate to `phpunit.xml.dist` structure/formatting/config.

I am not immediately certain how we can make our own internal test suite runnable on 10 and >10. It looks like the `doctrine-test-bundle` created two different `phpunit.xml` files, one for <10 and one for 10, so maybe there is a solution somehow: https://github.com/dmaicher/doctrine-test-bundle/tree/master/tests

In short, I'm not sure what you are asking here... do you want me to revert the changes to `phpunit.xml.dist` and then roll back all the PHPUnit 10 specific changes to our internal test files (`DataProvider` attribute, etc)? Or so I need to figure out how our internal testsuite can pass both PHPUnit8/9 and also PHPUnit10?

.gitignore
arderyp: done: https://github.com/arderyp/panther/commit/7124b9e85133b04cda0ead906b47024f56c0fea4

.gitignore
arderyp: thanks for the tip, done: https://github.com/arderyp/panther/commit/7124b9e85133b04cda0ead906b47024f56c0fea4

.gitignore
arderyp: yes phpunit to changed from `.phpunit.result.cache` to `.phpunit.cache'.  The [9.5 documentation](https://docs.phpunit.de/en/9.5/configuration.html) mentions a `cacheResultFile` property that can be defined in `phpunit.xml` to set a custom cache file, but I do not see it in the [10 documentation](https://docs.phpunit.de/en/10.0/configuration.html), so it may be deprecated.

composer.json
arderyp: or maybe you don't even want it in the dev dependencies? If so, I can delete this and `php-cs-fixer` from dev dependencies, just let me know.

composer.json
arderyp: Thanks for the info.

I am inexperienced with GitHub workflows. Should I be looking into changing our github workflow somehow, or are you just saying "revert back to phpunit-bridge, which will allow Github workflow to work properly, and just manually add/remove phpunit for my own testing locally"?

I am swamped tomorrow but hopefully can carve out some time in the afternoon, otherwise I can look on Friday.

composer.json
dunglas: Yes please delete it. The bot uses PHP CS Fixer under the hood but we don't need to install it locally.

phpunit.xml.dist
dunglas: If it's not possible to use a PHPUnit configuration that is compatible with both PHPUnit 10 and previous versions, then yes having two files will be necessary.

composer.json
arderyp: I've deleted phpstan and php-cs-fixer from dev dependencies: https://github.com/arderyp/panther/commit/cdcd7e08b58d93bf10a74c67cc6dbbde097dee89

That said, this is a confusing decision to me. Presumably, we should all be running phpstan before committing and only commit non-failing code. If so, why not keep phpstan as a dev dependency? Otherwise devs need to know to manually install it, then run it, then remember to manually delete it, then do it all over again for each commit. Furthermore, having it as a dev dependency doesn't hurt anything (or maybe I'm wrong?).

Likewise, on my initial PR, one of the Github actions failed for fixer related issues. If contributors are supposed to run the fixer themselves manually before committing, then it's the same weird expectation on dev to know they need to install it manually, run it, then remove it before commit, then do it all over again for the next commit.

@dunglas resolve if we don't need to run fixer/phpstan manually before committing (or we do and you still think they shouldn't be a dev dependencies for some reason)

phpunit.xml.dist
arderyp: The two files thing isn't really the problem, that's the easy part.  The potential problem is having internal tests that are compatible with 10 and <10.  But doctrine-test-bundle seems to be doing it, so I will give it a go and report back.

dunglas: Before wasting your time, I would like to hear what's @nicolas-grekas opinion on this.

Do we need to do so much efforts to support old PHPUnit versions, or can we expect a PHPUnit Bridge version compatible with PHPUnit soon and just release a new major?

arderyp: actually, may not be worth wasting _his_ time.  I spoke before testing.  I think I have this working. Commit incoming...

composer.json
arderyp: done: https://github.com/arderyp/panther/commit/cdcd7e08b58d93bf10a74c67cc6dbbde097dee89

@dunglas please resolve if sufficient

phpstan.neon
arderyp: done: https://github.com/arderyp/panther/commit/3c0c3f6f7d300bacdd4c1ce5fd42a10018edf07f

@dunglas please resolve if sufficient

phpunit.xml.dist
arderyp: done: https://github.com/arderyp/panther/commit/3c0c3f6f7d300bacdd4c1ce5fd42a10018edf07f

there is a new `phpunit.xml.dist.10` file for usage in PHPUnit 10 GitHub Action

@dunglas please resolve if sufficient

src/PantherTestCaseTrait.php
arderyp: I've removed this new method, but the remaining `\is_callable()` to `method_exists()` change remains to solve 2 phpstan errors.

@dunglas please resolve if sufficient

arderyp:
okay @dunglas, lots of reversions in https://github.com/arderyp/panther/commit/3c0c3f6f7d300bacdd4c1ce5fd42a10018edf07f which reduced the change set substantially.

I think I've addressed all of your comments above.  It would be helpful to de-clutter the space if you could just review all of my latest comments and resolve everything that you consider sufficient.

arderyp:
hey @dunglas, I feel like we had some traction last week and managed to get some good work done.  I think this is mostly ready to go, once you review.  I'd love to get back on it, if you've got the time :)

src/ServerExtension.php
```diff
@@ -13,86 +13,121 @@
...
 use PHPUnit\Runner\AfterTestHook;
 use PHPUnit\Runner\BeforeFirstTestHook;
 use PHPUnit\Runner\BeforeTestHook;
+use PHPUnit\Runner\Extension\Extension;
+use PHPUnit\Runner\Extension\Facade;
+use PHPUnit\Runner\Extension\ParameterCollection;
+use PHPUnit\TextUI\Configuration\Configuration;
 
-/**
+/*
```
dunglas, line 39:
```suggestion
/**
```

dunglas (approved):
LGTM

arderyp:
@dunglas I committed your suggestions, but the change to `src/ServerExtension.php` broke the CS clearance (I'm using the `.php-cs-fixer.dist.php` that ships on `main`).  also, there's something wonky with the `@phpstan-ignore-line` declarations in `PantherTestCaseTrait`.  I fixed those here to resolve the GitHub action errors: 

https://github.com/arderyp/panther/commit/e508f0c2d68878ffc3c28afb6b0121f504c49ab9
https://github.com/arderyp/panther/commit/4bca85670180904c42815455dd5e64acb7aac248

are we good to merge and release? Do the GitHub actions need to be updated to test phpunit 10 using the workaround you proposed [here](https://github.com/symfony/panther/pull/589#discussion_r1122382986)? It looks like they are still just running `9.5.28` via `composer install`. If this change is needed, is it something I can do, or do you need to edit the GitHub actions?

dunglas:
Yes, please create new GHA workflow to run the test with PHPUnit 10, then we'll able to merge.

@symfony/mergers could you take a look?

arderyp:
@dunglas I will look at this next week. Should I be duplicating eact phpunit workflow in `.github/workflows/ci.yaml` for full OS coverage, but with php `8.1` (minimum requirement for phpunit 10? or do you just want a single additional phpunit 10 / php 8.1 / ubuntu?

dunglas:
Just running the tests on 8.1 with PHPUnit 10 is good enough.

arderyp:
great. on ubuntu image (asopposed to windows/mac)?  basically copying the first `phphunit` actions, but with phpunit 10 and php 8.1?

arderyp:
@dunglas great. on ubuntu image (as opposed to windows/mac)?  basically copying the first `phphunit` [action](https://github.com/symfony/panther/blob/main/.github/workflows/ci.yml#L57-L89), but with phpunit 10 and php 8.1?

dunglas:
Yes

arderyp:
@dunglas @symfony/mergers I think we are good to go here :smiley: 

php8.1/phpunit10 test coverage added to new GitHub Action Workflow here (and passing): https://github.com/arderyp/panther/commit/48de6395697b07f724e6b8e56e25d7ca3027d47e

arderyp:
hey @dunglas, this has been approved for over a week.  What's the next step to get it merged?

shakaran:
@dunglas I would appreciate a lot if this PR is merged soon for update several github actions pipelines to PHPUnit 10. Thanks!

arderyp:
hey @dunglas. It's been many weeks.  How can we get the ball rolling on a merge? Who can I @ to get you what you need?

arderyp:
@dunglas @derrabus @fabpot @nicolas-grekas @stof 

This PR has been approved for 2 months.  How can we get it merged and released?

dunglas:
Thank you @arderyp, sorry for the long delay.

arderyp:
THANK YOU!

Is there any chance we will get a new release soon that includes this?

dunglas:
Yes! I'm on it.

dunglas:
Released!

PR #592, opened 2023-04-27
Improve DX when using the symfony binary

nicolas-grekas:
This PR improves the DX of using Panther when using the Symfony binary.

It improves the following:
- fix merging the base URI with the requested URL, adding/removing a slash between them as needed
- make the external URI default to `SYMFONY_PROJECT_DEFAULT_ROUTE_URL` when it's set
- auto-detect firefox when it's installed and chromium isn't
- ~accept insecure certs by default for firefox~

I spotted those when practicing the Symfony book:
https://symfony.com/doc/6.2/the-fast-track/en/17-tests.html

### DIFF
--- a/src/PantherTestCaseTrait.php
+++ b/src/PantherTestCaseTrait.php
@@ -55,7 +55,6 @@ trait PantherTestCaseTrait
         'router' => '',
         'external_base_uri' => null,
         'readinessPath' => '',
-        'browser' => PantherTestCase::CHROME,
         'env' => [],
     ];
 
@@ -100,7 +99,7 @@ trait PantherTestCaseTrait
             return;
         }
 
-        if ($externalBaseUri = $options['external_base_uri'] ?? $_SERVER['PANTHER_EXTERNAL_BASE_URI'] ?? self::$defaultOptions['external_base_uri']) {
+        if ($externalBaseUri = $options['external_base_uri'] ?? self::$defaultOptions['external_base_uri'] ?? $_SERVER['PANTHER_EXTERNAL_BASE_URI'] ?? $_SERVER['SYMFONY_PROJECT_DEFAULT_ROUTE_URL'] ?? null) {
             self::$baseUri = $externalBaseUri;
 
             return;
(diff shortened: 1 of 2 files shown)


src/PantherTestCaseTrait.php
```diff
@@ -162,10 +161,22 @@ protected static function createPantherClient(array $options = [], array $kernel
...
+        if (PantherTestCase::FIREFOX === $browser) {
             self::$pantherClients[0] = self::$pantherClient = Client::createFirefoxClient(null, null, $managerOptions, self::$baseUri);
+        } else {
+            try {
+                self::$pantherClients[0] = self::$pantherClient = Client::createChromeClient(null, null, $managerOptions, self::$baseUri);
+            } catch (\RuntimeException $e) {
+                if (PantherTestCase::CHROME === $browser) {
+                    throw $e;
+                }
+                $managerOptions['capabilities']['acceptInsecureCerts'] ??= true;
```
nicolas-grekas, line 173: I'm not sure this is the proper location for this. Should we could move it to the `Client` class?

I added this line because while the "Symfony dev CA" cert authority is installed for the regular firefox profile, it's not in headless mode, and I don't know why nor if it's possible to change that.

(I didn't check this aspect with Chrome yet.)

nicolas-grekas: Now tested on Chrome, which works like a charm.

src/ProcessManager/FirefoxManager.php
```diff
@@ -118,7 +118,9 @@ private function getDefaultOptions(): array
             'host' => '127.0.0.1',
             'port' => 4444,
             'path' => '/status',
-            'capabilities' => [],
+            'capabilities' => [
+                'acceptInsecureCerts' => true,
```
nicolas-grekas, line 122: Required on firefix headless because the "Symfony dev CA" isn't recognized there :shrug:

dunglas (approved):
LGTM, even if I don't like disabling certificate validation.

src/ProcessManager/FirefoxManager.php
nicolas-grekas: OK, removed. The rest is ready.

dunglas:
Thanks @nicolas-grekas!

Issue #598, opened 2023-06-18
No screenshot taken


vhb-Mark:
Hi all,

I have created a clean Symfony 6 application with phpunit and panther test.
Now I have the weird issue no screenshot is taken on a failed test?

Now I have found this part of the `PantherTestCaseTrait` and debugged it, it seems the `BaseTestRunner` class is not loaded?
![afbeelding](https://github.com/symfony/panther/assets/71921725/b2adc0dc-568f-4bc5-8bdd-7424fef6e5af)

Somebody have an option how I can fix this?

This is my code at the moment

`namespace App\Tests\Panther\Portal\User;

use Symfony\Component\Panther\PantherTestCase;

class UserLoginTest extends PantherTestCase
{
    public function testUserLoginSuccess(): void
    {
        $client = $this->createPantherClient();

        $client->waitForVisibility('#username', 5);
        $client->waitForVisibility('#password');

        $client->submitForm('btnLogin', ['username' => 'demo@test.com', '1234!Test']);
        $client->waitForVisibility('#menuTabDashboard');
        $this->assertSelectorIsVisible('#menuTabDashboard');
    }
}`

mathieu-dumoutier:
I confirmed that error screenshoot don't work since v2.1.0. When I force to 2.0.1, it's work.

PR #600, opened 2023-07-07
fix: no screenshot taken on test failure

norival:
See #598


dunglas (approved):

dunglas:
Good catch, thanks Xavier!

PR #604, opened 2023-09-26
Allow Symfony 7.0

kbond:


Chris53897:
Maybe it is time to bump minimum version from 5.3 to 5.4 (LTS)? Maybe in an seperate PR.
https://symfony.com/releases

dunglas:
LGTM but we need to make Psalm happy before merging. Maybe should we run the workflow with Symfony 7?

kbond:
> Maybe should we run the workflow with Symfony 7?

I did this - CI's happy now.

dunglas (approved):

dunglas:
Thank you!

PR #606, opened 2023-10-02
Fix example for api-platform

symac:
The wording has changed on the remote site. the current example does not work anymore as Get has been replaced by Getting


dunglas (approved):

dunglas:
Good catch, thanks!

Issue #615, opened 2023-12-04
TypeError is thrown when calling Crawler::html()


syl20b:
Hello, 

I'm upgrading from 2.1.0 to 2.1.1 and I encounter the following error : 

`TypeError: Symfony\Component\Panther\DomCrawler\Crawler::html(): Return value must be of type string, null returned`

It seems that the Crawler::html()  method must return a string, but as it calls the Crawler::attr() method under the hood, this can lead to a return null statement when the content is empty.

alexandre-mace:
I'm also having the same error, temporary downgraded to 2.1.0 (my code works again now) waiting for the PR to be merged, thanks for opening this issue, you saved me a bit of time @syl20b

PR #616, opened 2023-12-04
fix: allow content to be empty when calling Crawler::html() method

syl20b:
Hope that will fix the #615. 

Regards,

Sylvain

### DIFF
--- a/src/DomCrawler/Crawler.php
+++ b/src/DomCrawler/Crawler.php
@@ -229,7 +229,7 @@ final class Crawler extends BaseCrawler implements WebDriverElement
                 return $this->webDriver->getPageSource();
             }
 
-            return $this->attr('outerHTML');
+            return (string) $this->attr('outerHTML');
         } catch (\InvalidArgumentException $e) {
             if (null === $default) {
                 throw $e;
--- a/src/DomCrawler/Crawler.php
+++ b/src/DomCrawler/Crawler.php
@@ -229,13 +229,13 @@ final class Crawler extends BaseCrawler implements WebDriverElement
                 return $this->webDriver->getPageSource();
             }
 
-            return $this->attr('outerHTML', $default);
+            return $this->attr('outerHTML', (string) $default);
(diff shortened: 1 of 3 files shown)


syl20b:
@nicolas-grekas @dunglas is it ok for you?

dunglas (approved):

OskarStark (approved):

src/DomCrawler/Crawler.php
```diff
@@ -229,7 +229,7 @@ public function html(string $default = null): string
                 return $this->webDriver->getPageSource();
             }
 
-            return $this->attr('outerHTML');
+            return $this->attr('outerHTML', $default);
```
dunglas, line 232:
```suggestion
            return $this->attr('outerHTML', (string) $default);
```

Because we use strict types in Panther's code base.

dunglas (changes requested):

tests/DomCrawler/CrawlerTest.php
```diff
@@ -385,6 +385,15 @@ public function testHtmlDefault(callable $clientFactory): void
         $this->assertSame('default', $crawler->filter('header')->html('default'));
     }
 
+    /**
+     * @dataProvider clientFactoryProvider
+     */
+    public function testEmptyHtml(callable $clientFactory): void
+    {
+        $crawler = $this->request($clientFactory, '/basic.html');
```
dunglas, line 393: Could you also test with no arguments passed to `html()`?

dunglas (changes requested):

tests/DomCrawler/CrawlerTest.php
syl20b: @dunglas Done ;)

src/DomCrawler/Crawler.php
syl20b: @dunglas Done ;)

dunglas:
Thank you @syl20b!

PR #620, opened 2024-01-05
Document iterator type for Crawler

PrinsFrank:


dunglas (approved):

dunglas:
Thank you @PrinsFrank

PR #629, opened 2024-05-16
Green CI

HypeMC:
The CI is in a bit of a mess, hope this helps.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -153,6 +153,12 @@ jobs:
                   php-version: '8.3'
                   extensions: zip
 
+            - name: Install Firefox
+              run: brew install --cask firefox
+
+            - name: Install Geckodriver
+              run: brew install geckodriver
+
             - name: Install dependencies
               uses: ramsey/composer-install@v3
 


.github/workflows/ci.yml
```diff
@@ -194,7 +195,7 @@ jobs:
               run: vendor/bin/simple-phpunit
 
     phpunit-macos:
-        runs-on: macos-latest
+        runs-on: macos-13
```
HypeMC, line 198: macos-latest doesn't include the geckodriver

dunglas: If necessary, it is available on brew (which is preinstalled): https://formulae.brew.sh/formula/geckodriver

dunglas (approved):
Thank you!

.github/workflows/ci.yml
HypeMC: @dunglas Done, I had to install both Firefox & geckodriver, but it seems to be working now. However, it's kind of slow, over 15 min.

dunglas:
Thanks!

PR #164, opened 2019-01-17
Better support for non-standard webfolder

Nyholm:
This allows you to use config like: 
```xml
<php>
    <server name="PANTHER_WEB_SERVER_DIR" value="./web/" />
</php>
```

PR #632, opened 2024-06-12
Cast boolean env variables

SeyamMs:
Inspired by [#164](https://github.com/symfony/panther/issues/168)

Environment variables are always casted to strings, so added PHP `filter_var()` to cast `$_SERVER` variables at if statements.

I think for argument `PANTHER_DEVTOOLS` its required to be able to set it to `false`, while for arguments `PANTHER_NO_HEADLESS` and `PANTHER_NO_SANDBOX` are `false` by default and `true` if they just defined.

Hence, another applicable solution is to convert `PANTHER_DEVTOOLS` to `PANTHER_NO_DEVTOOLS`

### DIFF
--- a/src/ProcessManager/ChromeManager.php
+++ b/src/ProcessManager/ChromeManager.php
@@ -97,19 +97,19 @@ final class ChromeManager implements BrowserManagerInterface
         $args = [];
 
         // Enable the headless mode unless PANTHER_NO_HEADLESS is defined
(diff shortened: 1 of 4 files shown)


nicolas-grekas:
Can you please rebase so that we can trigger the CI again also?

OskarStark (approved):

dunglas (approved):

dunglas:
Thank you!

PR #635, opened 2024-07-25
[Feature] Add WaitForClickable helper

JoppeDC:


src/Client.php
```diff
@@ -507,6 +507,23 @@ public function waitForDisabled(string $locator, int $timeoutInSecond = 30, int
         return $this->crawler = $this->createCrawler();
     }
 
+    /**
+     * @param string $locator The path to the element that will be disabled. Can be a CSS selector or Xpath expression.
+     *
+     * @throws NoSuchElementException
+     * @throws TimeoutException
+     */
+    public function waitForClickable(string $locator, int $timeoutInSecond = 30, int $intervalInMillisecond = 250): PantherCrawler
```
javiereguiluz, line 516: `$timeoutInSecond` -> `$timeoutInSeconds` ?

same for `$intervalInMillisecond` -> `$intervalInMilliseconds` ?

JoppeDC: I thought this was a little weird too, but this is the way its done for all the other cases as well.

Cant change it for the other cases either, as thats a BC. We keep it like this so they're the same, or adjust this one>

javiereguiluz: I see. The same arguments in other methods are indeed named in the same way you did.

Changing the argument names can be made in other PR if wanted. By the way, Symfony's BC promise doesn't cover argument names, so we can change them if we decide it's better to do so. See https://symfony.com/doc/current/contributing/code/bc.html#note-10

Thanks

JoppeDC: Awesome, TIL!

I think its best to change all the names in a seperate PR in that case :)

dunglas:
The patch is empty, it looks like a rebase has gone wrong.

PR #641, opened 2024-10-02
Fix PHP 8.4 deprecations and CS

nicolas-grekas:

PR #642, opened 2024-10-02
Use own exception hierarchy

OskarStark:
Should be merged after
* https://github.com/symfony/panther/pull/641

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -39,6 +39,8 @@ use Symfony\Component\Panther\Cookie\CookieJar;
 use Symfony\Component\Panther\DomCrawler\Crawler as PantherCrawler;
 use Symfony\Component\Panther\DomCrawler\Form as PantherForm;
 use Symfony\Component\Panther\DomCrawler\Link as PantherLink;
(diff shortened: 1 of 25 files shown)


dunglas (approved):

nicolas-grekas:
rebase needed indeed :)

wouterj:
We used to say that Symfony components deserve their own exception hierarchy for runtime exceptions, but not for logic exceptions. The reason being that the latter are a logic error which should never be caught and must be fixed for the code to work at all.

OskarStark:
> We used to say that Symfony components deserve their own exception hierarchy for runtime exceptions, but not for logic exceptions. The reason being that the latter are a logic error which should never be caught and must be fixed for the code to work at all.

I double checked `symfony/symfony` before opening this PR and the `Asset`, `BrowserKit` and `Console` component uses exactly these ones: `RuntimeException`, `LogicException` and `InvalidArgumentException`, so I just followed this path 🤷‍♂️

OskarStark:
Will fix Tests on Monday

OskarStark:
Failures look unrelated

OskarStark:
friendly ping @dunglas

PR #651, opened 2024-12-10
feat: add PANTHER_REDUCED_MOTION to limit animations

Jean-Beru:
This PR add a `PANTHER_REDUCED_MOTION` environment variable to set the `prefers-reduced-motion` CSS media feature to `reduce`.

It could be useful to limit "flaky" tests which result in a `Facebook\WebDriver\Exception\ElementClickInterceptedException: Element <XXX> is not clickable at point (XXX, YYY) because another element XXX obscures it` exception because an animation is still in progress.

This CSS media feature is already supported by [Bootstrap](https://getbootstrap.com/docs/4.1/getting-started/accessibility/#reduced-motion) or the french [DSFR](https://github.com/GouvernementFR/dsfr).

### DIFF
--- a/src/ProcessManager/ChromeManager.php
+++ b/src/ProcessManager/ChromeManager.php
@@ -116,7 +116,7 @@ final class ChromeManager implements BrowserManagerInterface
         }
 
         // Prefer reduced motion, see https://developer.mozilla.org/fr/docs/Web/CSS/@media/prefers-reduced-motion
-        if ($_SERVER['PANTHER_REDUCED_MOTION'] ?? false) {
+        if (filter_var($_SERVER['PANTHER_REDUCED_MOTION'] ?? true, \FILTER_VALIDATE_BOOLEAN)) {
             $args[] = '--force-prefers-reduced-motion';
         }
 
--- a/src/ProcessManager/ChromeManager.php
+++ b/src/ProcessManager/ChromeManager.php
@@ -116,7 +116,7 @@ final class ChromeManager implements BrowserManagerInterface
         }
 
         // Prefer reduced motion, see https://developer.mozilla.org/fr/docs/Web/CSS/@media/prefers-reduced-motion
-        if (filter_var($_SERVER['PANTHER_REDUCED_MOTION'] ?? true, \FILTER_VALIDATE_BOOLEAN)) {
+        if (!filter_var($_SERVER['PANTHER_NO_REDUCED_MOTION'] ?? false, \FILTER_VALIDATE_BOOLEAN)) {
             $args[] = '--force-prefers-reduced-motion';
         }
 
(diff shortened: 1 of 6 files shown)


src/ProcessManager/ChromeManager.php
```diff
@@ -114,6 +114,11 @@ private function getDefaultArguments(): array
             $args[] = '--no-sandbox';
         }
 
+        // Prefer reduced motion, see https://developer.mozilla.org/fr/docs/Web/CSS/@media/prefers-reduced-motion
+        if ($_SERVER['PANTHER_REDUCED_MOTION'] ?? false) {
```
dunglas, line 118:
```suggestion
        if (filter_var($_SERVER['PANTHER_REDUCED_MOTION'] ?? true, \FILTER_VALIDATE_BOOLEAN)) {
```

(see #632)

src/ProcessManager/FirefoxManager.php
```diff
@@ -63,6 +64,14 @@ public function start(): WebDriver
         $capabilities = DesiredCapabilities::firefox();
         $capabilities->setCapability('moz:firefoxOptions', $firefoxOptions);
 
+        // Prefer reduced motion, see https://developer.mozilla.org/fr/docs/Web/CSS/@media/prefers-reduced-motion
+        if ($_SERVER['PANTHER_REDUCED_MOTION'] ?? false) {
```
dunglas, line 68:
```suggestion
        if (filter_var($_SERVER['PANTHER_REDUCED_MOTION'] ?? true, \FILTER_VALIDATE_BOOLEAN)) {
```

dunglas (changes requested):
Good idea! We should even enable this by default.

Jean-Beru:
> Good idea! We should even enable this by default.

I totally agree :slightly_smiling_face:

dunglas:
Maybe make this config the default and introduce a `PANTHER_NO_REDUCED_MOTION` to disable it then?

Jean-Beru:
It make sense with `PANTHER_NO_HEADLESS` :+1: 

I also added a test (and fixed a bug spotted by it).

src/ProcessManager/ChromeManager.php
```diff
@@ -115,6 +115,11 @@ private function getDefaultArguments(): array
             $args[] = '--no-sandbox';
         }
 
+        // Prefer reduced motion, see https://developer.mozilla.org/fr/docs/Web/CSS/@media/prefers-reduced-motion
```
dunglas, line 118:
```suggestion
        // Prefer reduced motion, see https://developer.mozilla.org/docs/Web/CSS/@media/prefers-reduced-motion
```

dunglas (approved):

dunglas:
Thank you!

Jean-Beru:
@dunglas FYI, I'm trying to understand why https://github.com/symfony/panther/actions/runs/12673280007/job/35319306027?pr=651 is failing

dunglas:
Thanks!

PR #652, opened 2024-12-18
feat: add explicit error messages in "waitFor*" methods

Jean-Beru:
Display an explicit message when a `waitFor*` assertion fails to make debug easier.


dunglas (approved):

dunglas:
Thanks!

PR #655, opened 2025-01-07
ci: Fixed static analysis

yohang:
Fixed some PHPStan errors and ignored some others (due to multiple dependencies support)

### DIFF
--- a/src/DomCrawler/Form.php
+++ b/src/DomCrawler/Form.php
@@ -211,10 +211,7 @@ final class Form extends BaseForm
         $this->setValue($field->getName(), $field->getValue());
     }
 
-    /**
-     * @return FormField|FormField[]|FormField[][]
-     */
-    public function get($name): FormField|array
+    public function get($name): FormField
     {
         return $this->getFormField($this->getFormElement($name));
     }
@@ -238,10 +235,8 @@ final class Form extends BaseForm
      * Gets the value of a field.
      *
      * @param string $name
-     *
-     * @return FormField|FormField[]|FormField[][]
      */
-    public function offsetGet($name): FormField|array
+    public function offsetGet($name): FormField
     {
         return $this->get($name);
     }
--- a/src/WebTestAssertionsTrait.php
+++ b/src/WebTestAssertionsTrait.php
@@ -39,8 +39,9 @@ trait WebTestAssertionsTrait
         $client = self::getClient();
 
         if ($client instanceof PantherClient) {
-            $element = self::findElement($selector);
-            self::assertNotNull($element, $message);
+            $by = $client::createWebDriverByFromLocator($selector);
+            $elements = $client->findElements($by);
+            self::assertNotEmpty($elements, $message);
 
             return;
         }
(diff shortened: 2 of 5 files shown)


src/DomCrawler/Form.php
```diff
@@ -211,10 +211,7 @@ public function set(FormField $field): void
         $this->setValue($field->getName(), $field->getValue());
     }
 
-    /**
-     * @return FormField|FormField[]|FormField[][]
-     */
-    public function get($name): FormField|array
```
dunglas, line 217: This looks like a BC break. Can't this method return an array anymore?

src/DomCrawler/Form.php
```diff
@@ -238,10 +235,8 @@ public function offsetExists($name): bool
      * Gets the value of a field.
      *
      * @param string $name
-     *
-     * @return FormField|FormField[]|FormField[][]
      */
-    public function offsetGet($name): FormField|array
```
dunglas, line 244: same

src/WebTestAssertionsTrait.php
```diff
@@ -39,8 +39,9 @@ public static function assertSelectorExists(string $selector, string $message =
         $client = self::getClient();
 
         if ($client instanceof PantherClient) {
-            $element = self::findElement($selector);
-            self::assertNotNull($element, $message);
+            $by = $client::createWebDriverByFromLocator($selector);
```
dunglas, line 42: Why this change?

tests/ScreenshotTest.php
```diff
@@ -37,7 +37,7 @@ public function testTakeScreenshot(): void
 
         $screen = $client->takeScreenshot();
 
-        $this->assertIsString($screen);
+        $this->assertNotEmpty($screen);
```
dunglas, line 40: Why this change?

dunglas (changes requested):
Thanks. To be honest I'm not confident with changing existing code just to make PHPStan happy. I would prefer ignoring the errors instead.

tests/ScreenshotTest.php
yohang: `$client->takeScreenshot()` has a string return type, so `$screen`  is a string in any cases. We just expect a method call without exception.

The `assertNotEmpty` is simply here to avoid the test to be reported as risky. I could add a `expectNotToPerformAssertions()` instead but it seemed ok to me to assert that a screenshot isn't an empty string.

src/WebTestAssertionsTrait.php
yohang: `findElement()` can't return null, it throws an `NoSuchElementException` if no element found, so the `assertNotNull` always pass. 

Instead, I check if that the return value of `findElements` isn't an empty array.

src/DomCrawler/Form.php
yohang: It seems that it can't, but I'm not fully confident about this one. I'll investigate a bit more.

yohang: Okay, I checked.

Base class (from `symfony/dom-crawler`) can return an array, this one can't as the private method `getFormField()` is already typed `FormField`.

It could be a BC break via inheritance of this class, but as it's `final`, it seems ok to me. What do you think ?

dunglas: LGTM

dunglas:
Thanks

PR #656, opened 2025-01-08
test: Added waitFor() calls in submit form tests to ensure crawler state

yohang:
`ClientTest::testSubmitForm()` is randomly failing. 

This PR adds a `waitFor()` call after form submits to fix the test.


dunglas (approved):
Thanks, I would like to find the underlying issue, but it's still better than a flaky CI.

PR #658, opened 2025-01-08
Fix reduced-motion tests on Windows

Jean-Beru:
Fix [tests on Windows](https://github.com/symfony/panther/actions/runs/12673985132/job/35321595051) introduced by #651.

It was due to the Windows image which enables reduced motion at OS level by default.


dunglas (approved):

dunglas:
Merci!

PR #679, opened 2025-08-25
Disable dev tools by default

### LINKED ISSUE #675: bug: `--auto-open-devtools-for-tabs` is broken on latest chromium version (137)
Hi,

I just wanted to report an annoying problem I encountered with the latest version of chromium, and how I managed to solve it (so that other developers don’t waste time on it).

# Bug
Using the version 137 of chromium, some requests will fail mysteriously with the following error code:

```json
{"code":-32000,"message":"Browser window not found"}
```

see https://stackoverflow.com/questions/79641152/error-code-32000-messagebrowser-window-not-found-with-usage-auto-op

# Solution
[The devtools are enabled by defaults on `symfony/panther`](https://github.com/symfony/panther/blob/b7e0f834c9046918972edb3dde2ecc4a20f6155e/src/ProcessManager/ChromeManager.php#L108-L110). I solved the problem either by using the previous version of Chromium (136), **which is not recommended for security reasons** if you’re accessing public websites, or by disabling DevTools with the PANTHER_DEVTOOLS=0 environment variable.

dmitryuk:
Closes #675

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+2.3.0
+-----
+
+* Disable dev-tools by default
+
 2.2.0
 -----
 


src/PantherTestCaseTrait.php
```diff
@@ -247,6 +248,7 @@ protected static function createHttpBrowserClient(array $options = [], array $ke
             self::$httpBrowserClient = new HttpBrowserClient(HttpClient::create($httpClientOptions));
         }
 
+        // @phpstan-ignore-next-line
         if (is_a(self::class, KernelTestCase::class, true)) {
             static::bootKernel($kernelOptions); // @phpstan-ignore-line
```
dmitryuk, line 253: Fixed failed CI

tests/ClientTest.php
```diff
@@ -282,8 +282,7 @@ public function testExecuteAsyncScript(): void
     const t = document.querySelector(parentArgs[0]).innerText;
     callback(t);
 }, 100, arguments);
-JS
-            , ['.p-1']);
+JS, ['.p-1']);
```
dmitryuk, line 285: Fixed failed cs-fixer in master

Kocal:
I think it's worth updating the changelog

PR #682, opened 2025-11-21
PHP 8.5 and Symfony 8 compatibility, remove EOLed versions

dunglas:

PR #678, opened 2025-08-05
Symfony 8 support follow-up, PHPUnit 11

Kocal:
IIRC the CI was already broken before this PR

Blockers:
- https://github.com/php-webdriver/php-webdriver/pull/1134

PR #686, opened 2025-12-01
Fixing Symfony 8 compatibility

ThomasLandauer:
I didn't test this!
But it looks like this was just forgotten in https://github.com/symfony/panther/pull/682#pullrequestreview-3522070506
It would also be included in https://github.com/symfony/panther/pull/678 but that is a larger PR that might take some more time.


Chris53897:
@ThomasLandauer I guess this can be closed in favor of the linked PR.

